PR #139, opened 2018-01-28
don't strip last CR-LF prior to terminating dot

fyfe:
Stripping the last \r\n off the DATA command results in the messages
actual size being 2 bytes shorter than the SIZE set in MAIL FROM.

RFC 1870[1]:
>   The message size is defined as the number of octets, including CR-LF
>   pairs, but not the SMTP DATA command's terminating dot or doubled
>   quoting dots, to be transmitted by the SMTP client after receiving
>   reply code 354 to the DATA command.

[1] https://tools.ietf.org/html/rfc1870#section-5

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -649,8 +649,6 @@ class SMTP(asyncio.StreamReaderProtocol):
                 self._writer.close()
                 raise
             if line == b'.\r\n':
-                if data:
-                    data[-1] = data[-1].rstrip(b'\r\n')
                 break
             num_bytes += len(line)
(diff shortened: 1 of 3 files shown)


waynew:
This is probably failing on Travis due to the IPv6 bug - a couple of the other PRs have added a fix that looks to fix that - you could either add that to this PR as well or wait for those to get merged.

waynew:
I went ahead and rebased this against master :+1:

waynew:
:disappointed: apparently I forgot to update my own upstream before rebasing. Doh!

Let's try that again...

Issue #123, opened 2017-07-07
LMTP server fails to list any extended services


tlhackque:
The LHLO service awaits smtp_HELO instead of smtp_EHLO, preventing the greeting from displaying the service menu.

This is a violation of the LMTP spec, RFC 2033.

Patch attached.  This affects https://gitlab.com/mailman/mailman/issues/365

[lmtp_lhlo.zip](https://github.com/aio-libs/aiosmtpd/files/1132381/lmtp_lhlo.zip)

Mortal:
```patch
--- /root/mailman3/venv-3.5/lib64/python3.5/site-packages/aiosmtpd/lmtp.py~	2017-07-05 19:50:44.000000000 -0400
+++ /root/mailman3/venv-3.5/lib64/python3.5/site-packages/aiosmtpd/lmtp.py	2017-07-07 16:51:35.488534153 -0400
@@ -5,11 +5,11 @@
 @public
 class LMTP(SMTP):
     @syntax('LHLO hostname')
     async def smtp_LHLO(self, arg):
         """The LMTP greeting, used instead of HELO/EHLO."""
-        await super().smtp_HELO(arg)
+        await super().smtp_EHLO(arg)
         self.show_smtp_greeting = False
 
     async def smtp_HELO(self, arg):
         """HELO is not a valid LMTP command."""
         await self.push('500 Error: command "HELO" not recognized')
```

PR #147, opened 2018-06-18
Fix LHLO service incorrectly waiting for smtp_HELO

justinclift:
Copied directly from Timothe Litt's fix in #123

### DIFF
--- a/aiosmtpd/lmtp.py
+++ b/aiosmtpd/lmtp.py
@@ -7,7 +7,7 @@ class LMTP(SMTP):
     @syntax('LHLO hostname')
     async def smtp_LHLO(self, arg):
         """The LMTP greeting, used instead of HELO/EHLO."""
-        await super().smtp_HELO(arg)
+        await super().smtp_EHLO(arg)
         self.show_smtp_greeting = False
 
     async def smtp_HELO(self, arg):


waynew:
Well, I guess when you get some tests we can merge ;)

justinclift:
Heh Heh Heh

Well, if @tlhackque has the right bits I can cut-n-paste, I'm happy to add them to this PR.

That being said, I probably should have done it properly first time around (eg create commit from the command line, used proper attribution on it, etc), but... just went went with the lazy approach instead and used GitHub's in-browser editor. :roll_eyes:

@tlhackque, if you do have a matching test case adjustment then I can get this PR done properly (eg so your name is on the commit instead of time).  Or just make a PR yourself and I'll close this one. :smile:

tlhackque:
Sorry - this came from my attempt to use Mailman V3, but I gave up as it wasn't/isn't ready for (my) prime time.  I no longer keep a current environment.  I don't think I updated the tests when I did; I don't know how your test system works.

I'm not a Python person; I was talking to the server from Perl, and not getting the correct responses.

I think the base story is pretty clear in my note from a year ago.

The test is probably failing because (a) it needs to send the extended HLO and (b) it will now get a list of  available services in response.  Some are mandatory; some depend on how the server is configured.

E.G. I just found the logs, and you can see that here:
````
File "C:\projects\aiosmtpd\aiosmtpd\tests\test_lmtp.py", line 28, in test_lhlo
    self.assertEqual(response, bytes(socket.getfqdn(), 'utf-8'))
AssertionError: b'APPVYR-WIN\nSIZE 33554432\n8BITMIME\nHELP' != b'APPVYR-WIN'
````
SIZE, HELP, and 8BITMIME are offered services; apparently the test is looking for a greeting.

I don't need credit - this was easy to find & fix and I was glad to contribute.

Eventually I hope that MM3 evolves to the point that I can try it again.

Thanks for getting around to working on it.

justinclift:
No worries.  I'll close this then, and whoever picks up the fix for #123 can use this for reference info. :smile:

PR #149, opened 2018-08-07
Replace in handle_DATA list of bytes into bytearray

ikrivosheev:
Hello, this is small memory optimization. My use case: I have an smtp server (aiosmtpd) that receives a 20-MB message stream. In current version I have two many copies of message:
1) data in handle_DATA - 20MB
2) content == original_content == 20MB
Sum is 40MB for one message. If we use an bytearray, we can save a little.

Thank you.

### DIFF
--- a/aiosmtpd/handlers.py
+++ b/aiosmtpd/handlers.py
@@ -60,7 +60,7 @@ class Debugging:
             if in_headers and not line:
                 print(_format_peer(peer), file=self.stream)
                 in_headers = False
-            if isinstance(data, bytes):
+            if isinstance(data, bytearray):
                 # Avoid spurious 'str on bytes instance' warning.
                 line = line.decode('utf-8', 'replace')
             print(line, file=self.stream)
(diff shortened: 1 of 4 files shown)


WolkoDav:
I do not really understand what went wrong...

waynew:
Hi  @ikrivosheev thanks for the PR!

I've been on a bit of a hiatus, but I'm back in the saddle now, and looking to get some stuff done!

How much space is saved when you use bytearray vs bytes? Have you done any measurement? Also I'm curious if there's any speed impact to doing this. If there's a significant improvement in either of those areas without a significant impact to the other, this could be a good approach.

Thanks again!

pepoluan:
-1

This has the _potential_ to mess up existing `handle_DATA` implementation that expected `bytes` instead of `bytearray`.

Unless one can set a flag requesting usage of `bytearray` instead of `bytes` ... but that increases complexity of the `smtp_DATA` method, which is already complex as it is.

Unless @ikrivosheev can show definite proof of significant space saving, I'm for rejecting this PR.

pepoluan:
It's one calendar month since my question and there is no response. I guess it's no longer necessary?

I'm going to close this commit for the time being.

Feel free to re-open if anyone thinks this is important.

PR #151, opened 2018-08-21
Update testing - python 3.7 and fix certs

scop:


waynew:
Weird, that seems like it should have been straightforward, but https://travis-ci.com/aio-libs/aiosmtpd/jobs/141108069#L727 is where the problem is. Maybe isn't called on 3.7 as well as Windows? That seems bizarre!

samuelcolvin:
Can I suggest using [codecov](https://codecov.io/) instead of the coverage checks in tox?

Given that python 3.7 is released it would be really good to have tests for 3.7.

waynew:
I have a mild preference for using coverage, mainly because it's a Python tool and it also seems to be the most widely used tool in the Python ecosystem.

samuelcolvin:
codecov is not an alternative to coverage, indeed it uses coverage. (cpython itself uses codecov).

It's just a site for showing the coverage. What I mean by "instead" is that codecov can "fail" a build if it's coverage drops so you can remove those checks tox and avoid problems like this.

waynew:
@samuelcolvin Oh, that makes a lot more sense then. I'll take a look into setting this up!

samuelcolvin:
@waynew did you have a look at switching to codecov? Would be great to get tests passing with python 3.7 and coverage to be easier to view.

waynew:
@samuelcolvin wow life got busy for me the past little while. Finally made some time to look into this... and by the time I finished writing this comment the codecov build passed :laughing:

If I understand correctly, this PR should be rebased from master now, to get the codecov stuff working properly?

samuelcolvin:
Yes. I think from memory (on my phone so can't look right now) you also need to stop tox or nose checking and failing on coverage changes.

scop:
Rebased.

waynew:
Well, I've finally made it back around these parts, and decided to try running the tests locally, only to discover that the certs were out of date.

So l went ahead and regenerated some new certs that don't expire for 100 years. I think that should be long enough :trollface:

Then the weirdest thing happened - I couldn't reproduce the test failure. At all. So I pushed the new certs and it looks like everyone and everything is happy here, so there we go. Merge time!

waynew:
p.s. thanks everyone for your help!

PR #156, opened 2018-12-07
Implemented global timeout and limited command call

cnicodeme:
This update adds a global timeout for the whole connection (on top of the timeout between each valid call) along with a counter for each command, to avoid malicious user for opening a connection and keeping it open by sending `NOOP` commands every few seconds.

It is an implementation based on the discussion in #145

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -76,6 +76,8 @@ class SMTP(asyncio.StreamReaderProtocol):
                  tls_context=None,
                  require_starttls=False,
                  timeout=300,
+                 global_timeout=300,
+                 command_call_limit=10,
                  loop=None):
         self.__ident__ = ident or __ident__
         self.loop = loop if loop else make_loop()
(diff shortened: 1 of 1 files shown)


cnicodeme:
(Just did a second push to fix a double underscore mistake)

waynew:
Hi @cnicodeme thanks for this PR! From what I can tell this looks :+1: - though we'll need some automated tests before we can merge this!

cnicodeme:
Of course, no problems on that!
(I don't understand what is the issue on the Travis checks)

waynew:
Hey @cnicodeme do you have any plans to update this PR with tests? Also it may be a good idea to rebase this against master.

pepoluan:
Command call limit is interesting... I think I can implement that in a different way, though.

Probably best to start a new PR.

pepoluan:
BTW, I've implemented Command Call Limit in [my branch](https://github.com/pepoluan/aiosmtpd/tree/command_call_limit).

My implementation maintains backward compatibility with current behavior (i.e., no command call limit).

It is not yet made into a PR because it depends on some new behavior as set by #213 .

pepoluan:
My PR #222 is live. The difference is that #222 focuses _only_ on limiting command call, because global timeout is already implemented by `SMTP._timeout_handler`, `SMTP._reset_timeout()`, and `SMTP._timeout_cb()`

So because both cases of this PR has been covered, I vote for closing this PR.

If no objections within 4x24 of this comment, I'll close this.

Issue #154, opened 2018-11-29
Hooks returning None should be treated as not procesed


cnicodeme:
When hooks are called for each steps of the SMTP call, the code check if the `MISSING` value is returned, which is an instance of `object()`.

I'm curious to know why.

From my perspective, if a call to a hook returns `None`, the code should continue like if there were no hooks at all.
The reason behind that is if no response is returned from the hook, the server still has to return a status, and the default one is accepted by the developer.

Moreover, if a hook exists, it MUST implement the logics presents in smtp.py when no hooks exists. Let me explain:

[Here's the line for `HELO` commands](https://github.com/aio-libs/aiosmtpd/blob/master/aiosmtpd/smtp.py#L338)

If there are no hooks, here's what's happening:

```
        status = await self._call_handler_hook('HELO', hostname)
        if status is MISSING:
            self.session.host_name = hostname
            status = '250 HELP'
```

But as soon as I add a `HELO` handler, the `status` won't be `MISSING`, which means I have to implement the two consequent lines in my own code, by **copy/pasting** it. That's not good.

Instead, if the smtp.py code does the following:
```
        status = await self._call_handler_hook('HELO', hostname)
        if status is None:
            self.session.host_name = hostname
            status = '250 HELP'
```

My handle can return `None`, saying "everything is good" like all hooks behave, and smtp.py will do the intended work.

But maybe I'm missing a specific reason, so I'd be curious to know.

Otherwise, happy to submit a PR if you want!

pepoluan:
Well, you can return `MISSING` from you handler. `MISSING` is after all an object importable from `aiosmtpd.smtp`

pepoluan:
Understood. I may have misconstrued what I meant to say in my comment there, my bad.

Changing the behavior might have extensive repercussions, though, for people who already got 'bitten' by this awkward interaction _and_ adapted, so any changes will need to be vetted & tested carefully.

Besides, there seems to be a plan to totally rewrite how EHLO handler works ( #155 ), and changes there can -- and should -- trickle down to the HELO handler.

cnicodeme:
Oh yes, ok, said like that it makes total sense!

I saw your suggestion on #155 and it's a good way to allow both `MISSING` and `None`, then put a DeprecationWarning in a few releases later, and then finally stop it? That way, there is no breaking updates :)

pepoluan:
> 
> 
> Oh yes, ok, said like that it makes total sense!
> 
[...]

PR #157, opened 2018-12-10
Allow handle_EHLO to change the responses to EHLO

### LINKED ISSUE #155: EHLO command returns all bunch of 250- data, even when the hooks returns otherwise
[In the current state of the code](https://github.com/aio-libs/aiosmtpd/blob/master/aiosmtpd/smtp.py#L351), some data are pushed BEFORE calling the hook, including

```
    250-{hostname}
    250-SIZE ...
    250-8BITMIME
    250-SMTPUTF8
    250-STARTTLS
```

And AFTER that, the hook is called.

The issue is that if the hook doesn't return a 2xx response, the server sends a mixed response and makes the clients fails.

The `EHLO` command should call the hook, and depending on the response, send all the push commands afterward.

cnicodeme:
Closes #155


aiosmtpd/smtp.py
```diff
@@ -352,28 +352,33 @@ def _set_rset_state(self):
...
         if hasattr(self, 'ehlo_hook'):
             warn('Use handler.handle_EHLO() instead of .ehlo_hook()',
                  DeprecationWarning)
             await self.ehlo_hook()
         status = await self._call_handler_hook('EHLO', hostname)
         if status is MISSING:
             self.session.host_name = hostname
-            status = '250 HELP'
-        await self.push(status)
+            response.append('250 HELP')
```
balki: For loop has to be outside the if block. It is now executed only when the hook is missing.

waynew: IIUC that's the point - if there's an EHLO hook, there should be a response - otherwise we should assume that the EHLO hook is going to specify all of the above. @cnicodeme is that correct?

waynew: Okay, yeah, refreshing myself with the issue at hand, yeah, that's the case, and this is definitely the right place for it.

aiosmtpd/smtp.py
```diff
@@ -352,28 +352,33 @@ def _set_rset_state(self):
...
+            response.append('250-SMTPUTF8')
             self.command_size_limits['MAIL'] += 10
         if self.tls_context and not self._tls_protocol:
-            await self.push('250-STARTTLS')
+            response.append('250-STARTTLS')
         if hasattr(self, 'ehlo_hook'):
             warn('Use handler.handle_EHLO() instead of .ehlo_hook()',
                  DeprecationWarning)
             await self.ehlo_hook()
         status = await self._call_handler_hook('EHLO', hostname)
```
waynew: I think that what I would like to see is an update to this call, maybe something like `ehlo_defaults` or something (actually, given our discussion on #155, I think it would make perfect sense to rename `response` to `ehlo_defaults` and then just pass it to the hook).

Technically I think that's changing up the API and probably needs a doc update as well, but that feels like the right thing to do for this case. Discussion is very welcome, though!

aiosmtpd/smtp.py
balki: yeah. got now. However not clear on the responsibilities. Whether the hook is supposed to push the status and other messages too. If so it cannot return any status.

waynew: The hook shouldn't be handling any communication, it should only be returning status lines that will then be passed to the client.

aiosmtpd/smtp.py
pepoluan: +1 on passing the list of responses to EHLO handler

aiosmtpd/smtp.py
```diff
@@ -352,28 +352,33 @@ async def smtp_EHLO(self, hostname):
...
         if status is MISSING:
```
pepoluan, line 375: While we're on the topic of rewriting `smtp_EHLO`, if we also consider #154 , what about changing this line to:

```python
if status is MISSING or status is None:
```

aiosmtpd/smtp.py
pepoluan: Accidentally deleted my earlier comment, ugh -_-

Retyping from memory: If we expect the hook to be "sane"-ly constructed (see #154 ) then the "if" there should be `if status is MISSING or status is None:`

If the hook wants to handle the communication -- after all, we passed `self` to it -- then it should return "something" (other than `MISSING` or `None`). In which case, the `else:` clause should not be there.

pepoluan:
Okay I'm really thinking about backward compatibility here...

_Current behavior_ is that `handle_EHLO` only _adds_ to existing list. This is the behavior of the code in `master`, and _it is documented_.

Changing this behavior _will_ break code that expects that behavior.

So I will implement the following strategy:

Depending on the **signature** of the `handle_EHLO` hook,
  * If `handle_EHLO` takes only 3 (three) args -- as is expected currently -- do the old behavior (adding)
  * If `handle_EHLO` takes > 3 args -- adding the list of responses as the 4th arg -- do the new behavior (replacing)

pepoluan:
Changing this to draft because there are some behavior that might break existing code.

Will undraft after I implement backward-compatibility workarounds.

pepoluan:
I will undraft this after testing on my test systems.

aiosmtpd/smtp.py
pepoluan: Resolved by the latest set of commits.

aiosmtpd/smtp.py
pepoluan: Resolved by redefining a completely new behavior for `handle_EHLO`.

This new behavior is implemented in a way that's backward-compatible, and it has been documented.

aiosmtpd/smtp.py
```diff
@@ -492,32 +512,53 @@ async def smtp_EHLO(self, hostname):
...
+            # REPLACE existing list of responses with what the hook returns.
+            # We will handle the push()ing
+            response.append('250 HELP')
+            response = await self._call_handler_hook("EHLO", hostname, response)
+
+        for r in response:
+            await self.push(r)
+
+        # Defensive. Why must Session.host_name be set by EHLO hook anyways?
+        if self.session.host_name is None:
```
waynew, line 560: *question*: would `if not self.session.hostname:` be preferred?

waynew (approved):
Just one question but not a blocker. I like the backwards compatibility here - I think we should plan on doing a deprecation/warning introduction release regularly, like maybe once per year or something. But that's orthogonal to this PR 🙃

aiosmtpd/smtp.py
pepoluan: Hmm, now that you mention it, makes sense. Apparently `check_helo_needed()` only checks for false-y state.

I originally thought that the HELO/EHLO handler can set the host_name attribute to non-None false-y value to signal further handlers down the pipeline of the need to do some special processing. But since `check_helo_needed()` does not check for None, then that possibility is nulled.

After a bit of thinking, I think I'll just remove that defensive check-and-set code; handle_EHLO / handle_HELO can simply let host_name be false-y to indicate that HELO/EHLO step failed.

pepoluan:
Currently doing testing on my test systems:

- [ ] Windows 10 (via PyCharm tox runner): `(ALL)`

**This comment is undergoing editing**

pepoluan:
> 
> 
> This pull request **introduces 1 alert** when merging [a0f963b](https://github.com/aio-libs/aiosmtpd/commit/a0f963b5df26b78136f0f5de7627c2319d44ab94) into [6c8fdbc](https://github.com/aio-libs/aiosmtpd/commit/6c8fdbc7aecabb66272bd906c123808048308885) - [view on LGTM.com](https://lgtm.com/projects/g/aio-libs/aiosmtpd/rev/pr-7e4af4db78fb0e1cc647c8090bc27d8f7316b1c9)
> 
> **new alerts:**
> 
>     * 1 for Unused import

Huh? I thought I've removed that line?

Oh well, here's to another edit-commit-push-check.

pepoluan:
[All my tests passed](https://github.com/aio-libs/aiosmtpd/pull/157#issuecomment-752350408), I'm Undrafting this.

pepoluan:
Whoops, there are some conflicts. Redrafting while I fix those.

pepoluan:
Because of the major merging, I'm rerunning my tests

Progress:
- [ ] Windows 10 (via PyCharm runner): `(ALL)`

**_This post is undergoing editing_**

pepoluan:
Ugh, finally FreeBSD testing is complete.

I don't know why but my FreeBSD VM seems to ... degrade? It usually finishes in an acceptable time without much fuss, but now it runs really slowly I had to add another core ... and I get intermittent failures due to timeouts.

I'll undraft this and open this for review.

aiosmtpd/tests/test_smtp.py
```diff
@@ -338,16 +338,17 @@ def test_ehlo(self):
...
                 b'SMTPUTF8',
                 b'AUTH LOGIN PLAIN',
                 b'HELP',
-            )
-            for actual, expected in zip(lines, expecteds):
-                self.assertEqual(actual, expected)
+            ]
+            # for actual, expected in zip(lines, expecteds):
+            #     self.assertEqual(actual, expected)
+            self.assertEqual(expecteds, actuals)
```
pepoluan, line 351: Better to compare whole-list against whole-list, to see where we went wrong.

tox.ini
```diff
@@ -74,7 +75,6 @@ deps:
     {[qadocs]deps}
 
 [flake8]
-exclude = conf.py
```
pepoluan, line 77: Do not exclude `conf.py` se we can catch possible problems.

tox.ini
```diff
@@ -54,6 +54,7 @@ passenv =
 basepython = python3
 envdir = {toxworkdir}/python3
 deps =
+    colorama
```
pepoluan, line 57: Small test-dep, really helpful for parsing the reams of output.

pepoluan:
If no objections within 2x24 I'll squamerge this 😄

pepoluan:
> 
> 
> This pull request **introduces 1 alert** when merging [a682b4a](https://github.com/aio-libs/aiosmtpd/commit/a682b4aac9594f698483f497428f6ab9827d955a) into [9432d76](https://github.com/aio-libs/aiosmtpd/commit/9432d7666dd265a62567163640d27607d426b999) - [view on LGTM.com](https://lgtm.com/projects/g/aio-libs/aiosmtpd/rev/pr-e0099feccab148f12edcc709da12a6fcac5c3c29)
> 
> **new alerts:**
> 
>     * 1 for Unused import

Oh fer cryin' out loud ...

pepoluan:
Carrying out final verification, just in case merging with master causes problems:

- [ ] Windows 10 (via PyCharm): `(ALL)`

**_This post is undergoing editing_**

pepoluan:
Marinating this for 2x24, then squamerge.

PR #171, opened 2019-10-04
Socket activation and systemd supervision

dkg:
Please consider these changes to enable eaiser integration of aiosmtpd implementations with systemd and other comparable systems.

### DIFF
--- a/aiosmtpd/main.py
+++ b/aiosmtpd/main.py
@@ -140,7 +140,7 @@ def main(args=None):
     if args.debug > 2:
         loop.set_debug(enabled=True)
 
-    if args.supervised:
+    if args.supervised:                           # pragma: nomswin
         log.info('Server listening on socket-activated file descriptor')
         sock = socket.fromfd(3, socket.AF_INET, socket.SOCK_STREAM)
         stage = loop.create_server(factory, sock=sock)
--- a/aiosmtpd/main.py
+++ b/aiosmtpd/main.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import signal
+import socket
 import asyncio
 import logging
 
@@ -61,7 +62,12 @@ def parseargs(args=None):
         '-d', '--debug',
         default=0, action='count',
         help="""Increase debugging output.""")
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        '--supervised',
+        default=False, action='store_true',
+        help="""Run supervised under systemd-style socket activation.""")
+    group.add_argument(
         '-l', '--listen', metavar='HOST:PORT',
         nargs='?', default=None,
         help="""Optional host and port to listen on.  If the PORT part is not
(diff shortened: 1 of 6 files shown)


dkg:
fwiw, i think the travis CI test is failing because the test wants to listen on port 8025 when something else (maybe another test) is already listening on it.  i don't think that failure is due to any of the changes proposed here.

waynew:
Hi @dkg, thanks for the contribution! Have you run these tests locally?

My experience is that there's no such thing as unrelated test failures. Either the tests are poorly written or the code has side effects that weren't accounted for. I'm thinking that might be the case here, as I haven't seen this error show up before... but I could be wrong!

dkg:
@waynew -- you were right, somehow i'd mixed up my local tests.  I've just pushed a fix (my changes to `Controller.__init__` were accidentally ignoring the `port` argument) and now `pytest-3` succeeds for me locally.

Looks like they succeed now, but the coverage tests are failing, i think because of the changes i've made.  I'm not sure how to increase the coverage effectively for these changes though -- can you help me figure that out?

aiosmtpd/main.py
```diff
@@ -134,10 +140,13 @@ def main(args=None):
     if args.debug > 2:
         loop.set_debug(enabled=True)
 
-    log.info('Server listening on %s:%s', args.host, args.port)
-
-    server = loop.run_until_complete(
-        loop.create_server(factory, host=args.host, port=args.port))
+    if args.supervised:
+        log.info('Server listening on socket-activated file descriptor')
+        stage = loop.create_server(factory, sock=socket.fromfd(3, socket.AF_INET, socket.SOCK_STREAM))
```
waynew: It looks like these are the lines that aren't actually getting tested, according to what I can glean from Travis.

waynew: https://travis-ci.com/aio-libs/aiosmtpd/jobs/248519180#L566

dkg:
I've just tried to add tests to exercise that code, but i'm not sure whether they are what you want.  please let me know if you'd rather they were done differently!

waynew:
@dkg I think that fits with the existing pattern, so it seems reasonable to me.

Looks like there are a couple of more problems though:

https://travis-ci.com/aio-libs/aiosmtpd/jobs/248965906#L487
and
https://travis-ci.com/aio-libs/aiosmtpd/jobs/248965906#L466

dkg:
OK, flake complaints are all actually fixed now, but i'm stumped on how to get this additional coverage, because it looks like it should already be covered to me.  If you can help me understand the coverage test suite better, i'd be happy to take pointers.

dkg:
@waynew, any hints ?

waynew:
The overlap is because there are two branches there - neither is getting exercised.

Which tests should be exercising that code? I'm not seeing anything...

dkg:
if 39-44 is not getting exercised at all, then surely none of the tests within it are being exercised.  so either i would expect *no* subtests  to be listed, or *both* subtests to be listed.  but instead, only the 42-44 subtest is listed, and the 40-41 subtest is not listed.  This makes me doubt my understanding of the report :/

Anyway, the two cases identified are in `Controller.run` where:

 * 39-44: `self.sock is None`
 * 42-44: `self.hostname is None`

But `aiosmtpd/tests/test_server.py` contains `test_socket_error`, `test_server_attribute`, and `test_smtp_utf8`, each of which instantiates a `Controller` object with both `hostname` and `sock` left unset (thereby defaulting to `None`), which are then `start`ed or otherwise indirectly `run`, right?

Sorry if i'm asking obvious questions -- feel free to point me to answers even if they're things that you think i should already know.

waynew:
> Sorry if i'm asking obvious questions -- feel free to point me to answers even if they're things that you think i should already know.

You're doing fine 🙂

I'm not super familiar with those tests myself. What I would do is run the tests I'm expecting to hit that code, and perhaps throw `assert self.sock is None`. You should see that fail, given the coverage report.

pepoluan:
This is an old PR, more than 6 months old, only brought up recently due to CLAassistant.

Is there desire to continue with this PR? If not, I will close this.

pepoluan:
**Note to anyone wanting to implement similar functionality,** either by picking up this PR or submitting a new one:

I'm **-1** on changing `Controller` class. Especially considering the complexity added by #213 when it gets merged.

If `systemd`/`supervised` support is needed, IMHO we should create a new subclass of `Controller`, say, `SupervisedController`. In that class, `_testconn()` can (and should) be overridden to cater for the different method of activation when using `supervised`.

Adding a new class will also make merging much easier :-)

pepoluan:
Okay, no activities for more than 3 weeks. I'm closing this PR for now.

If anyone wants to pick this up, feel free to reopen.

PR #189, opened 2020-07-04
Python 3.9 support

moggers87:
## What do these changes do?

- Add Python 3.8 and 3.9 to test matrix
- `asyncio.Task.all_tasks` moved to `asyncio.all_tasks` in Python 3.7

## Related issue number

Fixes #188

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

PR #169, opened 2019-06-28
Use MagicMock to make sure create_server doesn't use AsyncMock on Python 3.8

tirkarthi:
## What do these changes do?

On Python 3.8 mocking async functions return an AsyncMock which used to be a MagicMock. Restore the older behavior to fix the tests in 3.8

## Are there changes in behavior for the user?

No, this change is tests only.

## Related issue number

#167 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] Add a new news fragment into the `CHANGES` folder

I am not sure these changes need to be added to CHANGES or docs. The project also needs to be updated to test on 3.8. If someone can help or suggest changes for travis and tox config it would be helpful. Thanks.

PR #190, opened 2020-07-04
Simplify Travis configuration and fix issue with coverage failing

moggers87:
## What do these changes do?

- Add Python 3.8 to test matrix
- Don't run *-nocov, tests are already run by *-cov
- Don't run *-diffcov, this is already covered by codecov

No changes for the user

## Related issue number

This should fix the coverage issues experienced in #169 and #189 

## Checklist

- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,18 @@
 language: python
+python:
+  - "3.5"
+  - "3.6"
+  - "3.7"
+  - "3.8"
 install:
   - pip install tox
   - python3 setup.py egg_info
   - pip install -r aiosmtpd.egg-info/requires.txt
-matrix:
-  include:
-    - python: "3.5"
-      env: INTERP=py35 PYTHONASYNCIODEBUG=1
-    - python: "3.6"
-      env: INTERP=py36 PYTHONASYNCIODEBUG=1
-    - python: "3.7"
-      env: INTERP=py37 PYTHONASYNCIODEBUG=1
-      dist: xenial
+env: INTERP=py${TRAVIS_PYTHON_VERSION//./} PYTHONASYNCIODEBUG=1
 before_script:
   # Disable IPv6. Ref travis-ci/travis-ci#8711
   - echo 0 | sudo tee /proc/sys/net/ipv6/conf/all/disable_ipv6
-script:
-  - tox -e $INTERP-nocov,$INTERP-cov,qa,docs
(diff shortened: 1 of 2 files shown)


moggers87:
Apparently not, still getting that same coverage failure on Python 3.8. I'm going to force-push a couple of times to get Travis to rerun the build. I have a suspicion that this coverage issue is intermittent

moggers87:
And the build passes without my actually changing anything :cat:

pepoluan:
Hi @moggers87 ,

> Apparently not, still getting that same coverage failure on Python 3.8. I'm going to force-push a couple of times to get Travis to rerun the build. I have a suspicion that this coverage issue is intermittent

Was the coverage issue on these lines in smtp.py:

```python
        if self.session.ssl is not None:            # pragma: nomswin
            # If STARTTLS was issued, return False, because True has no effect
            # on an SSL transport and raises a warning. Our superclass has no
            # way of knowing we switched to SSL so it might return True.
            #
            # This entire method seems not to be called during any of the
            # starttls tests on Windows.  I don't really know why, but it
            # causes these lines to fail coverage, hence the `nomswin` pragma
            # above.
            return False
```

Because I'm having intermittent coverage problems right there in my PR #192 

If so, I'm _severely_ tempted to just make that into a `pragma: nocover` line instead of just `mswin`

moggers87:
I don't remember but it's an intermittent issue (as seen by the build passing last time it was run) so I wouldn't want to add `nocover` to any lines to "fix" it. There's clearly a problem with the tests and/or coverage.

pepoluan:
In general, I like the changes. Can you merge/rebase from the latest master? I need to comment on some things and would rather do it on a diff against latest master.

pepoluan:
I'd say with the release of Python 3.9 on Travis CI, the dropping of Python 3.5, and general 'acceptance' of the annoying intermittent Travis CI failure, this PR is sadly falling behind.

So I'm going to close this PR. Feel free to reopen if this can be brought more up-to-speed.

Issue #188, opened 2020-07-04
asyncio.Task.all_tasks has been removed in Python 3.9


tirkarthi:
https://github.com/aio-libs/aiosmtpd/blob/ec804d9038989b20d2068ab3dc69ff0888a48257/aiosmtpd/controller.py#L64

Ref : https://github.com/python/cpython/pull/20874

moggers87:
The fix is quite simple:

```
try:
    _all_tasks = asyncio.all_tasks
except AttributeError:
    # For Python < 3.7
    _all_tasks = asyncio.Task.all_tasks
```

I did have a PR open with this patch, but Python 3.9 hasn't been released yet.

tirkarthi:
Python 3.9 will have it's last beta release by this month. The final release schedule is at https://www.python.org/dev/peps/pep-0596/

pepoluan:
Python 3.9.0 is released. Maybe time to revisit this?

moggers87:
It's not available on Travis just yet.

pepoluan:
> It's not available on Travis just yet.

True, but we can already test against 3.9 on our systems. On Linuxes, with the help of `pyenv` I've successfully tested against 3.9

pepoluan:
This has been implemented when we merged #192, so I'm closing this issue.

Feel free to reopen if necessary.

Issue #178, opened 2020-01-04
Controller does not pass host name to the protocol


agronholm:
The default implementation of `connection.factory()` does not pass the `hostname` argument to `SMTP()`. Because of this, `SMTP.__init__()` looks up the FQDN of the socket which, at least on my system, takes several seconds **during which the event loop is blocked**. Such a blocking operation should never be executed in the event loop thread.

waynew:
Thanks for the report @agronholm!

Looks like the failure is on this line here:

https://github.com/aio-libs/aiosmtpd/blob/ec804d9038989b20d2068ab3dc69ff0888a48257/aiosmtpd/controller.py#L32

agronholm:
This is the real problem though: https://github.com/aio-libs/aiosmtpd/blob/master/aiosmtpd/smtp.py#L94

I stumbled onto this because in my test runs, it always took suspiciously long for the SMTP server to respond. I traced the problem to the line above. I would recommend using `socket.gethostname()` instead if it has to happen in `SMTP.__init__()`. At least on my system, it gives the same result but returns instantly, unlike `socket.getfqdn()`.

waynew:
That's probably because you don't have a resolvable FQDN - https://stackoverflow.com/a/13932042/344286

I'm pretty sure that simply passing the hostname through should do the correct thing here.

v7a:
I had the same performance issues as Alex and fixed it by inheriting from the controller and overriding `factory()`:
```py
def factory(self):
    return Server(self.handler, hostname=self.hostname, enable_SMTPUTF8=self.enable_SMTPUTF8)
```
Pretty much the same is present in https://github.com/cole/aiosmtplib, in which the `SMTPConnection` class queries the FQDN when issuing a `helo` command. This can be fixed by querying the FQDN once beforehand and passing it as `source_address` upon construction `SMTPConnection` or a subclass thereof (e.g. `aiosmtplib.smtp.SMTP`).

pepoluan:
IMHO the solution suggested by @v7a is correct. If we already pass an explicit hostname to `Controller`, we should pass that as well to `SMTP` in `factory`.

I'm tempted to put this into PR #192 , but that PR is already so unwieldy so I'll defer until the PR is merged, then start a new PR to fix this.

pepoluan:
I accidentally implemented a way of passing through parameters to the `SMTP` class in 7289e6d because a need arose...

It's part of that honkin' huge PR #192 , though 😅

[...]

Issue #167, opened 2019-06-23
Test failures and DeprecationWarnings in Python 3.8


tirkarthi:
There are test case failures in Python 3.8 . I think this is due to mocking `create_server` which is now `AsyncMock` instead of `MagicMock` with changes in https://bugs.python.org/issue26467 where coroutines when patched return `AsyncMock`. Thus there are warnings due to the mock not being awaited. The old behavior can be used with `MagicMock` explicitly passed while patching. There also seems to be a bug that `AsyncMock` is not recording `call_args` and other values for synchronous API.

There is also DeprecationWarning since `StreamReader` and `StreamWriter` are merged into `Stream` class in https://bugs.python.org/issue36889. Attached is a log with 3.9.0a0 but these are reproducible on 3.8.0b1 as well.

[aiosmptd.log](https://github.com/aio-libs/aiosmtpd/files/3318345/aiosmptd.log)

waynew:
There's also `PendingDeprecationWarning: Task.all_tasks() is deprecated, use asyncio.all_tasks() instead`

pepoluan:
This will be irrelevant when #192 is merged because that PR took out the whole suite of test cases that caused these warnings.

PR #192, opened 2020-10-05
Implement AUTH LOGIN and AUTH PLAIN -- with tests

pepoluan:
## What do these changes do?

**This PR obsoletes #10 and #66**

### MAJOR changes:

* Implement SMTP AUTH with the LOGIN and PLAIN mechanisms.
* Implement extensible AUTH mechanism via the "handler hook" system
* Returns Enhanced Status Code (RFC3463 and RFC5248) for many situations (not yet exhaustive -- focusing mostly on SMTP AUTH and impacted commands)
* Some commands now emits error if AUTH is required but has not happened

### Minor changes:

* No longer targets Python 3.5 because a hard dependency (`atpublic`) requires Python >= 3.6
* Smattering of type hints here there and everywhere

Existing usage should not be impacted.

Users that implemented #66 must implement some changes, though. Namely:

* Change in `SMTP.__init__` kwarg names
* Change in authenticator 'callback' signature

## Major:

#102 

### Minor / related / accidental:

#133 #167 #178 #188 #194 

## Checklist

- [x] Add a new news fragment into the `CHANGES` folder


pepoluan (commented):
I have confidence in this code. I'd love a **code review**, and if anything is not okay, pray tell.

pepoluan:
Still couldn't find this mythical "CHANGES folder", so I just edited `NEWS.rst` directly.

pepoluan:
> 
> 
> This pull request **fixes 1 alert** when merging [68e006b](https://github.com/aio-libs/aiosmtpd/commit/68e006b4894617c205477054e90e6399f289bd9d) into [97730f3](https://github.com/aio-libs/aiosmtpd/commit/97730f37f4a283b3da3fa3dbf30dd925695fea69) - [view on LGTM.com](https://lgtm.com/projects/g/aio-libs/aiosmtpd/rev/pr-58f41e2a97505beaaea30420010ce175f1fa166a)
> 
> **fixed alerts:**
> 
>     * 1 for Module is imported with 'import' and 'import from'

I seriously have no idea why fixing this Semmle/LGTM alert caused coverage to drop ... but _ONLY_ for Python 3.7.

I do not see the same behavior on my laptop, which runs Windows. And yes, I have run the full tox test set several times.

pepoluan:
Heh, apparently `asyncio` is durn slow on PyPy. It already exhibited that behavior running tox on my WSL. But the test result of Travis confirms it.

I'll remove PyPy from `.travis.yml`.

The still niggling question is why coverage on Python 3.7 on Travis keep "oscillating" between 99% and 100%; and the issue is _always_ between **lines 233~242**.

pepoluan:
PYPY Y U HAZ STRANGE DIR STRUCTURE

pepoluan:
**First of all, about that Travis "Fail"**

Checking the log, it's coverage issue. For building against PyPy3. Which is strange because **all other builds have no coverage problems**.

And since first time my PR went through Travis, if there's coverage problem (outside of strange dir structures -- I'm looking at you PyPy!), it's always this line:

```python
if self.session.ssl is not None:
    return False
```

At this point I'm 99.99% sure that the coverage failure is due to external circumstances, not because of inherent test case inadequacy. So, my suggestion is to **ignore this failure**.

You can check [these screenshots for proof that coverage is 100% when I test](https://imgur.com/a/K3yUc3V) (Imgur album).

I'm done for now. I look forward to any code reviews.

pepoluan:
Gonna close & reopen to trigger travis.

pepoluan:
Let's trigger Travis. I'm 80% certain this time all test passes.

pepoluan:
See? It passes with flying colors! I didn't do anything but close & reopen this PR.

Let's get that Code Review coming!

tox.ini
```diff
@@ -50,8 +51,7 @@ deps:
     sphinx
 
 [flake8]
-enable-extensions = U4
 exclude = conf.py
-hang-closing = True
 jobs = 1
 max-line-length = 79
+ignore = E123, E133, W503, W504
```
pepoluan, line 57: Reason for ignoring all these:

`E123` -- this is a too-narrow interpretation of [PEP8's "closing brace"](https://www.python.org/dev/peps/pep-0008/#:~:text=The%20closing%20brace) guide.
`E133` -- likewise
`W503` & `W504` -- these are exclusive to one another. If you fulfill W503, you'll violate W504, and vice versa. Also both of them are too-narrow interpretation of [PEP8's "Line break before/after binary operator"](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

(Note, the most recent recommendation is actually `W504`, but that causes a LOT of `qa` issues. Let's fix that in a separate PR)

unittest.cfg
```diff
@@ -1,13 +1,19 @@
 [unittest]
 verbose = 2
-plugins = flufl.testing.nose
+plugins =
+    nose2.plugins.prof
```
pepoluan, line 4: `flufl.testing.nose` conflicts with `nose2.plugins.prof`

Because IMHO profiling is more useful than style guide, I sacrificed the former for the latter.

unittest.cfg
```diff
@@ -1,13 +1,19 @@
...
 
 [log-capture]
 always-on = False
+clear-handlers = True
+date-format = None
+; format = %%(name)s: %%(levelname)s: %%(message)s
+format = %%(asctime)s: %%(name)s: %%(levelname)s: %%(message)s
+filter = -nose
+log-level = DEBUG
```
pepoluan, line 13: Actually I'm still failing in capturing the logs so far. But these lines shouldn't mess up with the rest of the tests.

unittest.cfg
```diff
@@ -1,13 +1,19 @@
...
 
-[flufl.testing]
-always-on = True
-package = aiosmtpd
-setup = aiosmtpd.testing.helpers.setup
-teardown = aiosmtpd.testing.helpers.teardown
-start_run = aiosmtpd.testing.helpers.start
+[profiler]
+always-on = False
+filename = _dynamic/cPROFILE
```
pepoluan, line 17: Combined with the renaming Python script in tox.ini, this allows per-testenv cProfile data.

Also shoving them into `_dynamic` so can easily be cleaned up by removing that single directory.

aiosmtpd/smtp.py
```diff
@@ -19,7 +32,22 @@
...
+MISSING = _Missing()
+
+
+AuthMechanismType = Callable[["SMTP", List[str]], Awaitable[Any]]
+_TriStateType = Union[None, _Missing, bytes]
+
+
+class _AuthMechAttr(NamedTuple):
+    method: AuthMechanismType
+    is_builtin: bool
```
pepoluan, line 50: Some acrobatics to get type hinting to work nicely.

aiosmtpd/smtp.py
```diff
@@ -108,6 +145,35 @@ def __init__(self, handler,
...
+        self._handle_hooks: Dict[str, Callable] = {
+            m.replace("handle_", ""): getattr(handler, m)
+            for m in dir(handler)
+            if m.startswith("handle_")
+        }
+        self._smtp_methods: Dict[str, Any] = {
+            m.replace("smtp_", ""): getattr(self, m)
+            for m in dir(self)
+            if m.startswith("smtp_")
+        }
```
pepoluan, line 176: Embedded in this block of changes is my attempt to make the code more performant: Instead of keep invoking O(n) `getattr` every time, let's just cache all those `getattr` results in an O(1) dict.

aiosmtpd/smtp.py
```diff
@@ -239,36 +305,29 @@ async def _handle_client(self):
...
+                line: bytes = await self._reader.readline()
                 log.debug('_handle_client readline: %s', line)
                 # XXX this rstrip may not completely preserve old behavior.
                 line = line.rstrip(b'\r\n')
                 log.info('%r Data: %s', self.session.peer, line)
                 if not line:
                     await self.push('500 Error: bad syntax')
                     continue
-                i = line.find(b' ')
+                command, _, arg = line.partition(b" ")
```
pepoluan, line 316: `str.partition()` and `bytes.partition()` are amazingly fast and result in simpler follow-up code. Let's use them more :-)

Edit: The reason I replaced this was actually an attempt to make `_handle_client` faster. Because in PyPy3 all these for unknown reasons run slower, and enabling PYTHONASYNCIODEBUG flag points me that this method is one of the main reasons for the slowness. This speeds up `_handle_client` enough to greatly reduce the number of "slow callback" warning when testing against PyPy3

aiosmtpd/smtp.py
```diff
@@ -449,17 +657,21 @@ def _getparams(self, params):
         return result
 
     def _syntax_available(self, method):
-        if getattr(method, '__smtp_syntax__', None) is None:
+        if not hasattr(method, '__smtp_syntax__'):
```
pepoluan, line 660: Since we're not using the result of this particular `getattr()` anyways, why not just use `hasattr()`?

aiosmtpd/smtp.py
```diff
@@ -469,18 +681,19 @@ async def smtp_HELP(self, arg):
...
             code = 501
         commands = []
-        for name in dir(self):
-            if not name.startswith('smtp_'):
-                continue
-            method = getattr(self, name)
+        for name, method in self._smtp_methods.items():
             if self._syntax_available(method):
-                commands.append(name.lstrip('smtp_'))
+                commands.append(name)
```
pepoluan, line 686: As you can see, caching all those `getattr`s results in a simpler code. Should also be much faster because we can drop the two `str` methods there.

aiosmtpd/smtp.py
```diff
@@ -584,9 +805,6 @@ async def smtp_RCPT(self, arg):
             await self.push(syntaxerr)
             return
         address, params = self._getaddr(arg)
-        if address is None:
-            await self.push(syntaxerr)
-            return
```
pepoluan, line 589: `not address` already implictly includes `address is None`. Since there is no need to differ if address is None or bool, let's just merge the two into one. Should make `smtp_RCPT` a tad faster.

aiosmtpd/smtp.py
```diff
@@ -663,9 +885,8 @@ async def smtp_DATA(self, arg):
             return
         # Remove extraneous carriage returns and de-transparency
         # according to RFC 5321, Section 4.5.2.
-        for i in range(len(data)):
```
pepoluan, line 666: Whenever you see `range(len(o))`, 99% of the time it's replaceable with `enumerate(o)`.

aiosmtpd/smtp.py
```diff
@@ -663,9 +885,8 @@ async def smtp_DATA(self, arg):
...
-            text = data[i]
-            if text and text[:1] == b'.':
+        for i, text in enumerate(data):
+            if text.startswith(b'.'):
```
pepoluan, line 889: Slicing seems to be quite slow on PyPy3 (subjective feeling).

This is simpler, clearer, and faster. Three bool-ish ops + 1 slicing replaced by an optimized search.

aiosmtpd/tests/test_handlers.py
```diff
@@ -633,6 +654,10 @@ async def smtp_DATA(self, arg):
 
 
 class CapturingController(Controller):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.smtpd = None
+
```
pepoluan, line 660: See commit message for bca53500

aiosmtpd/tests/test_handlers.py
```diff
@@ -671,6 +696,10 @@ async def rset_hook(self):
 
 
 class DeprecatedHookController(Controller):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.smtpd = None
+
```
pepoluan, line 702: See commit message for bca53500

aiosmtpd/tests/test_smtp.py
```diff
@@ -7,8 +7,15 @@
...
-from aiosmtpd.smtp import SMTP as Server, __ident__ as GREETING
-from aiosmtpd.testing.helpers import reset_connection
+from aiosmtpd.smtp import MISSING, SMTP as Server, __ident__ as GREETING
+from aiosmtpd.testing.helpers import (
+    SUPPORTED_COMMANDS_NOTLS,
+    assert_auth_invalid,
+    assert_auth_required,
+    assert_auth_success,
+    reset_connection,
+)
```
pepoluan, line 17: These are centralized because they are often used, and if the output ever changes it will be much easier to change the expected output in one place rather than searching this honking big file for all instances of same.

aiosmtpd/tests/test_smtp.py
```diff
@@ -29,8 +89,17 @@ def factory(self):
...
 class TimeoutController(Controller):
+    Delay: float = 2.0
+
+    def factory(self):
+        return Server(self.handler, timeout=self.Delay)
+
+
+class RequiredAuthDecodingController(Controller):
     def factory(self):
-        return Server(self.handler, timeout=0.1)
```
pepoluan, line 33: Original timeout of `0.1` is too fast for Windows, causing intermittent fails.

By making this into a class attribute it allows easier tuning because changing this automatically adapts the failure case in `TestTimeout.test_timeout`

pepoluan (commented):
I have reviewed my code and left comments for some of the less-obvious changes.

aiosmtpd/controller.py
```diff
@@ -56,12 +64,17 @@ def start(self):
...
             raise self._thread_exception
 
     def _stop(self):
         self.loop.stop()
-        for task in asyncio.Task.all_tasks(self.loop):
+        try:
+            _all_tasks = asyncio.Task.all_tasks
+        except AttributeError:   # pragma: nocover
+            _all_tasks = asyncio.all_tasks
+        for task in _all_tasks(self.loop):
```
pepoluan, line 77: Similar to #189 

The `pragma: nocover` is _at the moment_ needed to get 100% coverage. I'm trying to implement [`coverage-conditional-plugin`](https://github.com/wemake-services/coverage-conditional-plugin/) but had a glitch with that plugin earlier today. When I fixed the issue with that plugin, I'll replace `nocover` with something better.

pepoluan (commented):
I've reviewed changes since I last commitpushed.

Additional comments on the changes have been added.

waynew:
Nice job! I'll take a look at this maybe Monday. I'm only on a phone until then

pepoluan:
*DON'T MERGE THIS YET* ... I have an incoming commit I'm still verifying on my computer. Shouldn't affect coverage or success of test cases, but very likely (I'm more than 90% confident here) solve an intermittent problem I've been experiencing during my test, namely `“Event loop stopped before Future completed.”` error.

waynew:
@pepoluan I converted this to a draft so it can't be merged. I *believe* it will still run all the checks in the interim, though.

aiosmtpd/main.py
```diff
@@ -134,23 +137,31 @@ def main(args=None):
...
+    server = server_loop = None
+    try:
+        server = loop.create_server(factory, host=args.host, port=args.port)
+        server_loop = loop.run_until_complete(server)
+    except RuntimeError as e:  # pragma: nocover
+        log.critical("\u001b[33;1m" + "#" * 600 + "\u001b[0m",
+                     exc_info=e)
+        raise
+    log.debug(f"server_loop = {server_loop}")
+    log.info('Server is listening on %s:%s', args.host, args.port)
```
pepoluan, line 150: This is the point where the error `Event loop stopped before Future completed.` _intermittently_ happened. The root cause was actually in `test_main.py`.

Wrapping some `log.debug` around the scene of the crime, plus a \*very\* eye-catching yellow flag ;-)

aiosmtpd/tests/test_main.py
```diff
@@ -116,11 +124,11 @@ def test_nosetuid(self):
...
         # may not be logging.ERROR.
-        log = logging.getLogger('mail.log')
-        default_level = log.getEffectiveLevel()
-        with patch.object(log, 'info'):
+        _log = logging.getLogger('mail.log')
+        default_level = _log.getEffectiveLevel()
+        with patch.object(_log, 'info'):
             main(('-n',))
-            self.assertEqual(log.getEffectiveLevel(), default_level)
+            self.assertEqual(_log.getEffectiveLevel(), default_level)
```
pepoluan, line 131: Variable renamed to suppress complaint of shadowing a higher-level `log` variable.

aiosmtpd/tests/test_main.py
```diff
@@ -52,7 +49,18 @@ def setUp(self):
...
+        # (i.e., create_server + run_until_complete + run_forever) *sometimes*
+        # takes more than 0.1 seconds, causing tests to fail intermittently
+        # with “Event loop stopped before Future completed.” error.
+        #
+        # Because the error is intermittent and infrequently happen (maybe
+        # only about 5-10% of testing attempts), I figure the actual time
+        # needed would be 0.1 +/- 20%; so raising this value by 900%
+        # *should* be enough. We can revisit this in the future if it needs
+        # to be longer.
+        loop.call_later(1.0, loop.stop)
```
pepoluan, line 63: This is the root cause of the intermittent `Event loop stopped before Future completed.` errors.

Raising the delay to `1.0` from `0.1` and testing the result like 30-40 times seems to indicate that the intermittent problem has been fixed.

aiosmtpd/main.py
```diff
@@ -134,10 +137,35 @@ def main(args=None):
...
+                server = None
+                if server_loop:
+                    loop.stop()
+                    loop.close()
+                    server_loop = None
+            if fails > 3:
+                log.critical("Too many failed attempts.")
+                raise
+        else:
+            break
```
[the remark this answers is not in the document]
pepoluan: This particular chunk of code has been removed in the final commit because it's no longer needed.

pepoluan (commented):
I have reviewed the notable changes between previous commit and this last commit.

setup.py
```diff
@@ -2,7 +2,7 @@
 from setuptools import setup, find_packages
 
 
-require_python(0x30400f0)
+require_python(0x30600f0)
```
pepoluan, line 5: We now have a hard dependency on Python >= 3.6

waynew:
Okay, so I didn't get to this earlier - things went a bit sideways for me at work 🙃 

I will carve out some time either tonight or tomorrow morning to look at this closer!

pepoluan:
> 
> 
> Okay, so I didn't get to this earlier - things went a bit sideways for me at work 🙃
> 
> I will carve out some time either tonight or tomorrow morning to look at this closer!

Feel free to criticize and/or cherry pick 👌

.gitignore
```diff
@@ -20,10 +20,13 @@ var/*
...
 .noseids
 docs/_build/
 nosetests.xml
 __pycache__/
-htmlcov
+htmlcov/
+_dynamic/
 coverage.xml
 diffcov.html
+.python-version
```
pepoluan, line 32: `pyenv` uses this file to limit available Python versions to a project

.gitignore
```diff
@@ -20,10 +20,13 @@ var/*
 .tox
 .DS_Store
 .coverage
+.coverage.*.*
```
pepoluan, line 23: Suppresses `git status` complaining of new files if tox is cancelled mid-test.

.gitignore
```diff
@@ -20,10 +20,13 @@ var/*
...
 .noseids
 docs/_build/
 nosetests.xml
 __pycache__/
-htmlcov
+htmlcov/
+_dynamic/
```
pepoluan, line 29: A directory to store dumps of possibly-useful info when running tox's testenvs (e.g., a dump of the currently-seen env vars)

aiosmtpd/smtp.py
```diff
@@ -417,6 +481,150 @@ async def smtp_STARTTLS(self, arg):
...
+        blob: bytes
+        await self.push(server_message)
+        line = await self._reader.readline()
+        blob = line.strip()
+        # '=' and '*' handling are in accordance with RFC4954
+        if blob == b"=":
+            log.debug("User responded with '='")
+            return None
+        if blob == b"*":
+            log.warning("User requested abort with '*'")
```
pepoluan, line 546: This is a `WARNING` instead of `INFO` or `DEBUG`, because aborting an `AUTH` mid-mechanism seems ... suspicious.

pepoluan:
Please note that some commits have been cherry-picked into #195 

I strongly recommend that before merging this, #195 should be merged first.

.gitignore
waynew: Would `.coverage*` be a better ignore?

.gitignore
waynew: 👍

aiosmtpd/controller.py
```diff
@@ -4,12 +4,14 @@
...
 from public import public
+from typing import Any, Dict
 
 
 @public
 class Controller:
     def __init__(self, handler, loop=None, hostname=None, port=8025, *,
-                 ready_timeout=1.0, enable_SMTPUTF8=True, ssl_context=None):
+                 ready_timeout=1.0, enable_SMTPUTF8=True, ssl_context=None,
+                 server_kwargs: Dict[str, Any] = None):
```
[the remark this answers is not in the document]
waynew: I *think* that seems reasonable!

aiosmtpd/controller.py
```diff
@@ -38,7 +42,11 @@ def _run(self, ready_event):
                 self.loop.create_server(
                     self.factory, host=self.hostname, port=self.port,
                     ssl=self.ssl_context))
-        except Exception as error:
+        except Exception as error:  # pragma: nowsl
+            # Somehow WSL1.0 (Windows Subsystem for Linux) allows multiple
+            # listeners on one port?!
+            # That is why we add "pragma: nowsl" there, so when testing on
+            # WSL we can specify "PLATFORM=wsl".
```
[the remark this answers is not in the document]
waynew: I'd believe it. That is definitely super weird, tho.

aiosmtpd/smtp.py
waynew: Can we check the timing of this, over:

```
try:
    return bool(getattr(self, method.__smtp_syntax_when__))
except AttributeError:
    return False
else:
    return True
```

waynew: If there's a significant difference we should do the faster one, but otherwise, I'm cool with this current change.

aiosmtpd/tests/test_main.py
waynew: if you change this to something super small like `0.01` can we change it from intermittent to always?

tox.ini
```diff
@@ -1,46 +1,47 @@
 [tox]
-envlist = {py35,py36,py37}-{cov,nocov,diffcov},qa,docs
+envlist = {py36,py37,py38}-{cov,nocov,diffcov}, qa, docs
 skip_missing_interpreters = True
 
 [testenv]
 commands =
+    python -c "import sys; print(sys.version); print(sys.executable)"
+    python -c "import os,pprint; os.makedirs('_dynamic', exist_ok=True); pprint.pprint(dict(os.environ), stream=open('_dynamic/ENV."{envname}"', 'wt'))"
```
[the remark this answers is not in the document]
waynew: Will this expose any types of credentials from Travis & co?

tox.ini
waynew: FWIW, I'd be happy to adopt `black` for code formatting, because auto formatting is 👍

waynew (changes requested):
I think this looks great! There are a couple of places where you noted that need some typo fixes. A couple of suggestions/ideas, but overall, 👍  love it!

aiosmtpd/controller.py
pepoluan: Good idea, I'm putting in the comment now, to go into the next commit round.

.gitignore
pepoluan: That will exclude `.coveragerc` as well, though. The default "partial coverage filename format" is, [according to coverage.py documentation](https://coverage.readthedocs.io/en/coverage-5.3/cmd.html#combining-data-files-coverage-combine), `.coverage.${HOSTNAME}.${PID}.${RANDOM}`.

The two periods there I think was back when the config file was still named `.coverage.ini`, so I had to use two periods to not exclude the config file.

If we want to simplify that exclusion, we still need to leave in at least one period so we won't accidentally exclude `.coveragerc`

aiosmtpd/main.py
```diff
@@ -146,11 +174,11 @@ def main(args=None):
     log.info('Starting asyncio loop')
     try:
         loop.run_forever()
-    except KeyboardInterrupt:
+    except KeyboardInterrupt:  # pragma: nocover
```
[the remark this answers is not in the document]
pepoluan: I also hate `pragma:nocover`, and actually I have a branch -- still in my laptop, not yet pushed to github -- where I'm dastardly trying to invoke a `KeyboardInterrupt` event.

It's really hard -- if not outright impossible, but I'm working on it -- to do in Windows, though.

So I consider that to be a "placeholder nocover", just like the one in the Python 3.9 compatibility shim (about the `asyncio.all_tasks`).

I'll put in a comment to explain the situation.

aiosmtpd/docs/NEWS.rst
```diff
@@ -5,9 +5,18 @@
...
+* Support for SMTP ``AUTH``, with AUTH hooks feature
+* Built-in implementation for ``AUTH PLAIN`` and ``AUTH LOGIN`` logic
+* Feature to inject keyword args during server class instantiation in ``Controller.factory``
+
 Fixed
 -----
 * Don't strip last ``\r\n`` prior to terminating dot.
+* A more maintainable test suite
+* Drop support for Python 3.5
+
```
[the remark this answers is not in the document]
pepoluan: It's not the docs that mentioned "CHANGES", actually. But the template when creating a new PR.

I see several other PR submitters also got dumbfounded by that checkbox 😁

aiosmtpd/smtp.py
```diff
@@ -108,6 +145,35 @@ def __init__(self, handler,
...
         self.transport = None
         self._handler_coroutine = None
+        if not auth_require_tls and auth_required:
+            warn("Requiring AUTH while not requiring TLS "
+                 "can lead to security vulnerabilities!")
+        self._auth_require_tls = auth_require_tls
+        if auth_callback is None:
+            self._auth_callback = login_always_fail
+        else:
+            self._auth_callback = auth_callback
```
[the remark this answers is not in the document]
pepoluan: Good idea! Why didn't I think of that...

Putting it in the next commit batch as well.

aiosmtpd/smtp.py
```diff
@@ -417,6 +481,150 @@ async def smtp_STARTTLS(self, arg):
...
+            return
+        args = arg.split(' ')
+        if len(args) > 2:
+            await self.push('501 Too many values')
+            return
+        mechanism = args[0]
+        if mechanism not in self._auth_methods:
+            await self.push('504 5.5.4 Unrecognized authentication type')
+            return
+        status = await self._call_handler_hook('AUTH', args)
```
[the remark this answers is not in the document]
pepoluan: We can do that for the first five `if`s. Afterwards the `if`s are responding to improper parameter contents, so we shouldn't put them into the `if/elif` chain.

aiosmtpd/smtp.py
```diff
@@ -417,6 +481,150 @@ async def smtp_STARTTLS(self, arg):
...
+    #      - If the client provides "=" for username during interaction, the
+    #        method MUST return b"" (empty bytes)
+    # 3. Auth credentials checking is performed in the auth_* methods because
+    #    more advanced auth mechanism might not return login+password pair
+    #    (see #2 above)
+
+    async def auth_PLAIN(self, _, args: List[str]):
+        loginpassword: _TriStateType
+        if len(args) == 1:
+            # Trailing space is MANDATORY
```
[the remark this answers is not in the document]
pepoluan: Good idea. I'm adding a comment in the code to go into the next commit batch.

aiosmtpd/smtp.py
pepoluan: Gut feeling says it's about the same 😁. But let me do some `timeit`s and I'll get back to you.

tox.ini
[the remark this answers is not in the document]
pepoluan: I don't think so ...

That line never show the infodump on-screen; all the dumps go into an 'internal' directory that gets `.gitignore`d and omitted from coverage.

Unless we can access the virtual machine where Travis do their tests, I don't think any Travis info gets leaked.

aiosmtpd/smtp.py
```diff
@@ -108,6 +145,35 @@ def __init__(self, handler,
...
+        self.authenticated = False
+        # Get hooks & methods to significantly speedup getattr's
+        self._auth_methods: Dict[str, _AuthMechAttr] = {
+            m.replace("auth_", ""): _AuthMechAttr(getattr(h, m), h is self)
+            for h in (self, handler)
+            for m in dir(h)
+            if m.startswith("auth_")
+        }
+        for m in (auth_exclude_mechanism or []):
+            self._auth_methods.pop(m, None)
```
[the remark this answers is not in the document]
pepoluan: Makes sense. Let me add a logging line there to list the 'found' mechanisms.

aiosmtpd/smtp.py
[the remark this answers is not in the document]
pepoluan: Okay, it doesn't result in "too much indentation", which I'm actively avoiding. So I'll implement the change in the next commit batch.

aiosmtpd/smtp.py
pepoluan: [Checkout this gist](https://gist.github.com/pepoluan/908996ede1da06c0058061db40d89016)

Surprisingly, the one with `hasattr` is nearly **twice as fast** as the one with `try ... except` 😲

Based on this result, I will use the current implementation with `hasattr`

aiosmtpd/tests/test_smtp.py
```diff
@@ -18,9 +25,62 @@
...
+        self.session = None
+
+    async def handle_MAIL(
+            self, server, session, envelope, address, mail_options
+    ):
+        self.session = session
+        return "250 OK"
+
+    async def auth_NULL(
+            self, server, kargs
```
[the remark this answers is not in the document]
pepoluan: Typo fixed in next commit batch.

tox.ini
```diff
@@ -1,46 +1,47 @@
 [tox]
-envlist = {py35,py36,py37}-{cov,nocov,diffcov},qa,docs
+envlist = {py36,py37,py38}-{cov,nocov,diffcov}, qa, docs
```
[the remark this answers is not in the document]
pepoluan: And this is a bit outdated now: I've added `py39` in the list. Works well on my three test platforms.

pepoluan:
I accidentally borked my repo. Please wait while I unbork it

pepoluan:
I don't understand why Github decided that it cannot merge 4 simple changes with the latest commit set, but I've checked the conflicts and all of them can be resoved by accepting the ones from `auth_base`

pepoluan (commented):
I believe all requested changes have been implemented, and those that I have implemented, I have marked as "Resolved"

pepoluan:
Gaaah, I'm almost 100% quite certain that it passed `qa` on my laptop /*headdesk/*

Okay will commit a `qa` fix.

.gitignore
pepoluan: This is merged in #195, so I'm marking as resolved.

.coveragerc
```diff
@@ -18,3 +18,4 @@ source =
 exclude_lines =
     pragma: nocover
     pragma: no${PLATFORM}
+    pragma: skipif_${PY_39}_py39
```
pepoluan, line 21: Custom semi-automatic exclusion of Python 3.9-related lines

aiosmtpd/controller.py
```diff
@@ -72,16 +72,7 @@ def _stop(self):
...
-            # I don't like using pragma:nocover just like everyone else, but at
-            # the time being this is the only way to achieve 100% coverage on
-            # 3.6 <= Python < 3.9. (One day *if* we drop support for 3.6, we
-            # can revisit and simplify this, because 3.7 implemented
-            # asyncio.all_tasks)
-            # That said, I'm planning to use coverage-conditional-plugin, but I
-            # have hit some problems that prevented me from using that plugin's
-            # 'conditional pragmas'. As soon as I can get that problem fixed,
-            # I'll replace the pragma:nocover.
+        except AttributeError:   # pragma: skipif_lt_py39
```
pepoluan, line 75: Yay! We've killed the `nocover` and replaced it with a better "conditional" pragma.

tox.ini
```diff
@@ -33,6 +33,8 @@ setenv =
     py39: INTERP=py39
     pypy3: INTERP=pypy3
     nocov: PYTHONASYNCIODEBUG=1
+    !py39: PY_39=lt
+    py39: PY_39=ge
```
pepoluan, line 37: These two lines work together to mark whether the Python interpreter is < 3.9 (`lt`) or >= 3.9 (`ge`); this env var will be used later in `.coveragerc`

aiosmtpd/docs/NEWS.rst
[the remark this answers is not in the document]
waynew: Ah. That template should be found in the `.github` folder - we should take some time to fix that!

waynew (approved):
🚀  Nice!

Issue #162, opened 2019-02-09
If the connection is broken, an exception will be generated


wevsty:
I'm having some problems with my code, check the systemd log as follows.
```
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]: Exception in handle_exception()
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]: Traceback (most recent call last):
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:   File "/usr/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 315, in _handle_client
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:     await method(arg)
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:   File "/usr/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 710, in smtp_DATA
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:     await self.push('250 OK' if status is MISSING else status)
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:   File "/usr/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 224, in push
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:     await self._writer.drain()
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:   File "/usr/lib/python3.6/asyncio/streams.py", line 339, in drain
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:     yield from self._protocol._drain_helper()
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:   File "/usr/lib/python3.6/asyncio/streams.py", line 210, in _drain_helper
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:     raise ConnectionResetError('Connection lost')
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]: ConnectionResetError: Connection lost
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]: During handling of the above exception, another exception occurred:
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]: Traceback (most recent call last):
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:   File "/usr/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 326, in _handle_client
Feb 06 15:21:23 hostname-vm start_filter.sh[20874]:     await self.push(status)
[...]

Issue #127, opened 2017-08-10
ConnectionResetError when client disconnects abruptly


sanjioh:
Hi,

when a client disconnects abruptly, `aiosmtpd` logs an unhandled `ConnectionResetError`, that follows an attempt to return a status code to the client once it has already gone.

I managed to reproduce the issue quite consistently with the following handler and CLI command:

```python
import asyncio
from concurrent.futures import CancelledError, ProcessPoolExecutor


def func():
    pass


class Handler:
    def __init__(self):
        self._executor = ProcessPoolExecutor(4)

    async def handle_DATA(self, server, session, envelope):
        loop = asyncio.get_event_loop()
        try:
            await loop.run_in_executor(self._executor, func)
        except CancelledError:
            print('cancelled')
        return '250 OK'
```

CLI: `python -m aiosmtpd -n -c handler.Handler`

I used a benchmarking tool (https://github.com/nabeken/go-smtp-source) as client, which sends a user defined number of messages of a certain size, concurrently.
If I stop the test hitting CTRL-C on the client, I get the following output from aiosmtpd (sometimes it takes a couple of retries to reproduce, but it should be quite easy):

```
cancelled
ERROR:mail.log:SMTP session exception
Traceback (most recent call last):
  File "/home/fabio/.virtualenvs/mailsampler/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 281, in _handle_client
    await method(arg)
  File "/home/fabio/.virtualenvs/mailsampler/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 677, in smtp_DATA
    await self.push('250 OK' if status is MISSING else status)
  File "/home/fabio/.virtualenvs/mailsampler/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 203, in push
    await self._writer.drain()
  File "/opt/pyenv/versions/3.6.2/lib/python3.6/asyncio/streams.py", line 323, in drain
    raise exc
  File "/opt/pyenv/versions/3.6.2/lib/python3.6/asyncio/selector_events.py", line 724, in _read_ready
    data = self._sock.recv(self.max_size)
ConnectionResetError: [Errno 104] Connection reset by peer
ERROR:mail.log:Exception in handle_exception()
Traceback (most recent call last):
  File "/home/fabio/.virtualenvs/mailsampler/lib/python3.6/site-packages/aiosmtpd/smtp.py", line 292, in _handle_client
    await self.push(status)
[...]

PR #193, opened 2020-10-11
Fix ConnectionResetError when the client disconnected

wevsty:
## What do these changes do?

Added error handling for ConnectionResetError.
This will improve the appearance of long exception messages.

Make sure that handle_exception() is not called when the client is actively disconnected.

## Related issue number

* Make sure that handle_exception() do not called when the client disconnect and cancelled. (issue #127 #162)
This will fix connection lost will print errors ConnectionResetError.

## Checklist

- [x] Add a new news in NEWS.rst


pepoluan:
Hi @wevsty , can you merge/rebase latest master into your branch so I can test it?

wevsty:
@pepoluan  
I've done rebase. For now you can test it.

aiosmtpd/tests/test_smtp.py
```diff
@@ -1386,6 +1397,20 @@ def test_exception_handler_exception(self):
...
+        controller.start()
+        self.addCleanup(controller.stop)
+        with SMTP(controller.hostname, controller.port) as client:
+            code, response = client.ehlo('example.com')
+            self.assertEqual(code, 250)
+            with self.assertRaises(SMTPServerDisconnected) as cm:
+                mail = CRLF.join(['Test', '.', 'mail'])
+                client.sendmail('anne@example.com', ['bart@example.com'], mail)
+            self.assertIsInstance(cm.exception, SMTPServerDisconnected)
+            self.assertEqual(handler.error, None)
```
pepoluan, line 1412: Good test. I have some suggestions:

1. After connection is disconnected, ensure that further attempts with `client` raises `SMTPServerDisconnected`. My suggestion:

```python
            # At this point connection should be down
            with self.assertRaises(SMTPServerDisconnected) as cm:
                client.mail("alice@example.com")
            self.assertEqual("please run connect() first", str(cm.exception))
```

2. Add another test case ensuring that existing connections are not impacted. You can have two `SMTP` objects, both already doing their HELO/EHLOs, then first object does a `.sendmail`, and after catching the exception, the second object does a `.mail` (that is, `MAIL FROM`) and this one should get a `250 OK`.

aiosmtpd/tests/test_smtp.py
```diff
@@ -175,6 +175,17 @@ async def handle_exception(self, error):
...
 
+class ErroringHandlerConnectionLost:
+    error = None
+
+    async def handle_DATA(self, server, session, envelope):
+        raise ConnectionResetError('ErroringHandlerConnectionLost test')
+
+    async def handle_exception(self, error):
+        self.error = error
+        raise ConnectionResetError('ErroringHandlerConnectionLost test')
```
pepoluan, line 186: Is raise necessary? I mean, at this point, you're just trying to ensure that `ErroringHandlerConnectionLost` is not called, why the `raise`?

aiosmtpd/smtp.py
```diff
@@ -380,7 +380,11 @@ async def _handle_client(self):
                 # verify that it was actually self._reader that was reset.
                 log.info('Connection lost during _handle_client()')
                 self._writer.close()
-                raise
+                break
```
pepoluan, line 383: I'm -1 on replacing `raise` with `break`

[Documentation since 3.7](https://docs.python.org/3.7/library/asyncio-exceptions.html#asyncio.CancelledError) clearly states that:

> In almost all situations the exception must be re-raised.

I'm not sure if our situation suits the small sliver of situation where we shouldn't re-raise.

In fact, testing the test cases on my computer, all tests still pass even if I revert `break` back to `raise`

aiosmtpd/docs/NEWS.rst
```diff
@@ -20,6 +20,8 @@ Fixed/Improved
 * No more failures/DeprecationWarnings for Python 3.8 (fixes #167)
 * Faster ``_handle_client()`` processing
 * Faster method access for ``smtp_*``, ``handle_*``, and ``auth_*`` hooks
+* Fix connection lost will print errors ConnectionResetError. (fixes #162)
+* Make sure that handle_exception() do not called when the client disconnected. (fixes #127)
```
pepoluan, line 24: Probably typo. Should have said:

`Make sure that handle_exception() do not get called when the client disconnected.`

aiosmtpd/docs/handlers.rst
```diff
@@ -121,6 +121,7 @@ synchronously (i.e. they are **not** coroutines).
     a status string, such as ``'542 Internal server error'``.  If the method
     returns ``None`` or raises an exception, an exception will be logged, and a
     ``500`` code will be returned to the client.
+    Note: If connection is lost function will not be called.
```
pepoluan, line 124: Please add an empty space, and strong-ify the "Note:" part, like this:

```rst
    ``500`` code will be returned to the client.

    **Note:** If connection is lost function will not be called.
```

pepoluan (changes requested):
I've reviewed the code, left some comments. There are some changes to make, and some points that I need some clarifications of.

aiosmtpd/tests/test_smtp.py
wevsty: I've combined the two suggestions together.
see the:
test_exception_handler_multiple_connections_lost

aiosmtpd/tests/test_smtp.py
wevsty: It's really not in use here. I've removed it.

aiosmtpd/tests/test_smtp.py
```diff
@@ -1386,6 +1396,30 @@ def test_exception_handler_exception(self):
...
+        self.addCleanup(controller.stop)
+        with SMTP(controller.hostname, controller.port) as client1:
+            code, response = client1.ehlo('example.com')
+            self.assertEqual(code, 250)
+            with SMTP(controller.hostname, controller.port) as client2:
+                code, response = client2.ehlo('example.com')
+                self.assertEqual(code, 250)
+                with self.assertRaises(SMTPServerDisconnected) as cm:
+                    mail = CRLF.join(['Test', '.', 'mail'])
+                    client2.sendmail('anne@example.com', ['bart@example.com'], mail)
```
pepoluan, line 1412: According to `tox -e qa` this line is too long. If you can chop it into two lines, it will pass Travis checking.

aiosmtpd/tests/test_smtp.py
```diff
@@ -1386,6 +1396,30 @@ def test_exception_handler_exception(self):
...
+                self.assertIsInstance(cm.exception, SMTPServerDisconnected)
+                self.assertEqual(handler.error, None)
+                # At this point connection should be down
+                with self.assertRaises(SMTPServerDisconnected) as cm:
+                    client2.mail("alice@example.com")
+                self.assertEqual("please run connect() first", str(cm.exception))
```
pepoluan, line 1418: According to `tox -e qa` this line is too long. If you can chop it into two lines, it will pass Travis checking.

aiosmtpd/docs/NEWS.rst
```diff
@@ -42,7 +44,7 @@ Removed
 1.1 (2017-07-06)
 ================
 * Drop support for Python 3.4.
-* As per RFC 5321, §4.1.4, multiple ``HELO`` / ``EHLO`` commands in the same
+* As per RFC 5321, ยง4.1.4, multiple ``HELO`` / ``EHLO`` commands in the same
```
pepoluan, line 47: Why is this changing?

aiosmtpd/docs/handlers.rst
```diff
@@ -89,7 +89,7 @@ The following hooks are currently defined:
     instantiated with ``decode_data=False`` (the default) or
     ``decode_data=True``.  In the former case, both ``envelope.content`` and
     ``envelope.original_content`` will be the content bytes (normalized
-    according to the transparency rules in |RFC 5321, §4.5.2|_).  In the latter
+    according to the transparency rules in |RFC 5321, ยง4.5.2|_).  In the latter
```
pepoluan, line 92: Why is this changing?

aiosmtpd/docs/handlers.rst
```diff
@@ -291,5 +292,5 @@ We open up the mailbox again, and all three messages are waiting for us.
 .. _Maildir: https://docs.python.org/3/library/mailbox.html#maildir
 .. _RFC 4954: https://tools.ietf.org/html/rfc4954
 .. |RFC 4954| replace:: **RFC 4954**
-.. _RFC 5321, §4.5.2: https://tools.ietf.org/html/rfc5321#section-4.5.2
-.. |RFC 5321, §4.5.2| replace:: **RFC 5321, §4.5.2**
+.. _RFC 5321, ยง4.5.2: https://tools.ietf.org/html/rfc5321#section-4.5.2
+.. |RFC 5321, ยง4.5.2| replace:: **RFC 5321, ยง4.5.2**
```
pepoluan, line 296: Why are these changing?

pepoluan (changes requested):
There are some changes that need to be checked. In summary:

* Section signs "§" got replaced; please revert these
* Some lines are too long, not passing `tox -e qa`, please fix these.

aiosmtpd/docs/handlers.rst
pepoluan: I prefer to have an empty line there, but I'll accept this.

aiosmtpd/smtp.py
wevsty: Here's where `raise` causes a problem.

asyncio.CancelledError was a subclass of Exception before Python 3.8.
So if use `raise`, error will be captured again.
But on the second capture it calls `self.push()` to write to the closed socket.
This is why issue #127 happened.

aiosmtpd/docs/NEWS.rst
wevsty: Changed

aiosmtpd/docs/handlers.rst
wevsty: Changed.

aiosmtpd/docs/NEWS.rst
wevsty: This is a coding error caused by the editor I'm using.
I'll fix it later.

wevsty:
@pepoluan 

>     * Section signs "§" got replaced with strange symbols; please revert these

I'm sorry that my editor coding caused some problems. 
The symbol "§" has been fully revert.
>     * Some lines are too long, not passing `tox -e qa`; please fix these.

Long lines I've made changes.

I see that Travis CI is showing a build error on py38.
```
py38-cov run-test: commands[5] | python -m coverage report -m --fail-under=100
Name                     Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------
aiosmtpd/__init__.py         0      0      0      0   100%
aiosmtpd/controller.py      55      0      6      0   100%
aiosmtpd/handlers.py       149      0     50      0   100%
aiosmtpd/lmtp.py            12      0      2      0   100%
aiosmtpd/main.py            89      0     18      0   100%
aiosmtpd/smtp.py           651      1    262      1    99%   246->255, 255
--------------------------------------------------------------------
TOTAL                      956      1    338      1    99%
```
My local test with Python 3.8.5 on ubuntu did not prompt an exception. 
The code comments also indicate that the code should skip the test.
I don't think that should be a problem.

pepoluan:
> My local test with Python 3.8.5 on ubuntu did not prompt an exception.
> The codes are as follows
> 
> ```python
>         if self.session.ssl is not None:            # pragma: nomswin
>             # If STARTTLS was issued, return False, because True has no effect
>             # on an SSL transport and raises a warning. Our superclass has no
>             # way of knowing we switched to SSL so it might return True.
>             #
>             # This entire method seems not to be called during any of the
>             # starttls tests on Windows.  I don't really know why, but it
>             # causes these lines to fail coverage, hence the `nomswin` pragma
>             # above.
>             return False
> ```
> 
> According to the code instructions, testing under win may not override.
> However, Travis CI gives the result of the uncovered test, which means that there is some probability that the test will not be covered on Linux as well.
> I don't think it's a problem, But replacing it here with `# pragma: no cover` might be a better choice.

It's kind of a well-known intermittent problem. _Very_ hard to nail down because it's not reproducible on any contributor's workstation. See #190 for example. I experienced it a LOT in #192.

Just close this PR and reopen it (since you're the PR's author, you can do both). That will trigger Travis to run again, hopefully the next time it runs, whatever caused that line to be skipped will not happen.

wevsty:
I've refreshed the CI,Now the build has passed.

aiosmtpd/smtp.py
pepoluan: What if we put in a flag there?

Something like this:

```python
# outside try block
was_cancelled_error = False
...
    except asyncio.CancelledError:
        ...
        was_cancelled_error = True
        raise
    ...
    except Exception as error:
        if was_cancelled_error:
            raise
        ...
```

Therefore we still follow the asyncio recommendation of re-raising `CancelledError`, but prevent the broad except from firing its contents.

pepoluan: That said, out of curiosity I ran this short snippet:

```python
class CustomException(Exception):
    pass


try:
    raise CustomException
except CustomException:
    raise
except Exception:
    print("recaught the exception")
```

And the `raise` within the first except block was never caught by the second block.

Tested on Ubuntu with Python 3.6, 3.7, and 3.8.

wevsty: I've done more testing and apparently you're right.

issue #127 should be the same as #162.
The error caused by calling `self.push()` with `handle_exception()` after the client disconnected.

Here I'll revert to `raise` and change the description.

pepoluan (approved):

pepoluan:
Sorry for the radio silence. My bandwidth is a bit full at the moment.

I'll do a test on latest of this PR on my systems later today. If they all pass, I'll bump the version and push. And if pass again by the PR checks, I'll merge.

pepoluan:
Test progress:
[ ] Windows 10 (via PyCharm runner): `(ALL)`

aiosmtpd/docs/NEWS.rst
```diff
@@ -21,6 +29,7 @@ Fixed/Improved
 * No more failures/DeprecationWarnings for Python 3.8 (Closes #167)
 * Faster ``_handle_client()`` processing
 * Faster method access for ``smtp_*``, ``handle_*``, and ``auth_*`` hooks
+* Make sure that handle_exception() do not get called when the client disconnected. (fixes #127 #162)
```
pepoluan, line 32: I could've sworn I _moved_ this line to up there...

BRB, committing & pushing.

pepoluan (commented):
Some ... unexpected things.

aiosmtpd/docs/NEWS.rst
```diff
@@ -2,6 +2,14 @@
  NEWS for aiosmtpd
 ===================
 
+1.2.3a1 (aiosmtpd-next)
```
pepoluan, line 5: Let's make this official

aiosmtpd/smtp.py
```diff
@@ -24,7 +24,7 @@
 from warnings import warn
 
 
-__version__ = '1.2.2'
+__version__ = '1.2.3a1'
```
pepoluan, line 27: It's official

pepoluan (approved):
I'm officially designating this as **1.2.3a1**

pepoluan:
After 24h if no objections from the other maintainers, we'll Squash & Merge this PR.

wevsty:
@pepoluan 
This looks good.
I don't have any questions.

PR #199, opened 2020-10-30
Release 1.2.2

waynew:
A version bump for uploading to PyPI

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -11,7 +11,7 @@ from public import public
 from warnings import warn
 
 
-__version__ = '1.2+'
+__version__ = '1.2.2'
 __ident__ = 'Python SMTP {}'.format(__version__)
 log = logging.getLogger('mail.log')
 


waynew:
@pepoluan Total sidenote - but opening this PR gave me the chance to check:

I'm pretty sure that the PR template comes from the aio-libs org and not this repo, as I don't find any `.github` folder. But that's also fine! I like the PR template. But... we should update this repo to use towncrier to generate the changelog from the CHANGES folder - it looks like the intent is that we're using more or less the keepachangelog.org format, which is a good thing. Anyway, we should probably create an issue for that?

pepoluan:
> But... we should update this repo to use towncrier to generate the changelog from the CHANGES folder - it looks like the intent is that we're using more or less the keepachangelog.org format, which is a good thing. Anyway, we should probably create an issue for that?

If we are going to use towncrier, then I'm all for that. We do need to document that, though. If not within the repo, maybe in the wiki?

pepoluan (approved):
No problem with this; let's get this merged! 🎉

waynew:
I think we should either add it to the readme, or create a HACKING text file... Oh! Yeah, since we include docs in the project, we should go ahead and add it there. It's probably a good idea to add a HACKING file too, though, that points to the docs.

PR #200, opened 2020-10-30
Release helpers

waynew:
This adds a release helper script - basically what I use to upload releases to pypi myself. I don't have any emotional investment in the configuration of this script - it's just me trying to improve that bus factor. If there are improvements then they are very welcome!

Anyway, the idea is that you should be able to create a pypi token and setup a `~/.pypirc`, install necessary packages (e.g. twine), and then you'd be able to run this script any time you needed to release!

In the process of setting up this script, the twine check informed me of the missing field, so I went ahead and fixed that.

### DIFF
--- a/release.py
+++ b/release.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+import sys
+
+import aiosmtpd.smtp as smtpd
+
+version = smtpd.__version__
+
+choice = input(f'Release aiosmtpd {version} - correct? [y/N]: ')
+if choice.lower() not in ('y', 'yes'):
+    sys.exit('Release aborted')
+else:
+    # We're probably already in the right place
+    os.chdir(os.path.dirname(os.path.abspath(__file__)))
+    # Let's use *this* python to build, please
+    subprocess.run([sys.executable, "setup.py", "sdist"])
+    # Assuming twine is installed. And that we're only building .tar.gz
+    subprocess.run(["twine", "check", f"dist/aiosmtpd-{version}.tar.gz"])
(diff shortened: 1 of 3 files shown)


pepoluan:
I'll take a look later today. Have to do some things this afternoon (I'm in UTC+7 so it's broad daylight as I type this 😁)

release.py
```diff
@@ -0,0 +1,28 @@
...
+    # Assuming twine is installed. And that we're only building .tar.gz
+    subprocess.run(["twine", "check", f"dist/aiosmtpd-{version}.tar.gz"])
+    # You should have an aiosmtpd bit setup in your ~/.pypirc - for twine
+    subprocess.run(["twine", "upload", "--config-file", "~/.pypirc", "-r", "aiosmtpd", "dist/aiosmptd-{version}.tar.gz"])
+    # Only tag when we've actually built and uploaded. If something goes wrong
+    # we may need the tag somewhere else!
+    # The annotation information should come from the changelog
+    subprocess.run(["git", "tag", "-a", version])
+    # And now push the tag, of course.
+    subprocess.run(["git", "push", "upstream", "--tags"])
```
pepoluan, line 28: If we're going to use towncrier, we'll need to update this script ;-)

pepoluan (approved):
The best scripts are scripts that allow the devs to be knowledgeably lazy (as opposed to ignorantly lazy) ... let's get this merged 🎉

release.py
waynew: Yup 😄 figured I'd at least move this forward - but I'd love to gen the tag annotation from a towncrier output. But... Future!

PR #211, opened 2020-11-22
GitHub Actions Workflow to replace Travis CI

pepoluan:
## What do these changes do?

Configures GitHub Actions, and stop using Travis CI.

Also banishes the very annoying & irritating "intermittent coverage failure" error that has been plaguing the PRs since, like, forever.

## Are there changes in behavior for the user?

None whatsoever. This is repo back-end only.

## Related issue number

Closes #209 

## Checklist

(I've removed irrelevant checklist items)

- [x] I think the code is well written

PR #202, opened 2020-11-08
Convert all test cases to pytest

pepoluan:
## What do these changes do?

Replace nose2 with pytest for testing.

## Are there changes in behavior for the user?

None. These are testing-side only.

## Related issue number

Closes #198 

Also partially answers #128 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist _-- well... these_ are _the unit tests..._
- [x] Unit tests have been run successfully for:
  - [x] Windows 10 (PyCharm runner): `(ALL)`
  - [x] Windows 10 (PSCore 7.0.3): `(ALL)`
  - [x] Windows 10 (Cygwin): `py36-{nocov,cov}` [1]
  - [x] FreeBSD 12.1 on VBox: `(ALL) + py36-nocov` [2]
  - [x] Ubuntu 18.04 on WSL 1.0: `(ALL) + pypy3-{nocov,cov,diffcov}`
  - [x] Ubuntu 18.04 on VBox: `(ALL) + pypy3-{nocov,cov,diffcov}`
  - [x] Ubuntu 20.04 on VBox: `(ALL) + pypy3-{nocov,cov,diffcov}`
- [x] Documentation reflects the changes
- [x] Update `aiosmtpd-next` section of NEWS.rst

[1] : Cannot install pyenv nor pypy3 on Cygwin, so using available Python only. diffcov also doesn't work.
[2] : pypy3 package on FreeBSD cannot run coverage.py nor pytest-cov; it complains of missing `_sqlite_cffi`.

PR #213, opened 2020-11-26
Catch exceptions raised by Controller.factory()

### LINKED ISSUE #212: Controller.factory() exceptions not caught
(Got bitten by this when mangling things up during unittest-to-pytest conversion)

As currently implemented, `Controller._run()` catches exceptions raised by `loop.create_server()`. The problem is that `create_server` is "lazy": It invokes the `factory()` method only when (1) a connection is made, _and_ (2) some data has been exchanged. Due to this "lazy"-ness, `create_server` only generates an Exception if it cannot bind to the `host:port` as specified.

As a result, if the call to `factory()` somehow caused an Exception, that Exception will go under the radar. And since `SMTP` hasn't been instantiated, `handle_exception` in the handler object won't ever be called. From client side, one ends up only with `smtplib.SMTPServerDisconnected: Connection unexpectedly closed` error, with no hints as to the root cause, potentially misleading users.

(I personally spent quite some time puzzling out why SMTPClient failed to connect, before finding out that I made a mistake in the instantiation of SMTP.)

Since users of this library has likely followed the example of subclassing `Controller` and overriding `factory()` with a simple `SMTP()` instantiation, I think the only non-disruptive backward-compatible way to implement a mechanism to catch `factory()` errors is to wrap calls to `factory()` using another (private) method whose sole purpose is to watch for success/failure of `factory()` and act accordingly.

I'm preparing a PR that will implement this mechanism.

**Edit:** The code is currently un-PR-ed in [pepoluan/aiosmtpd:catch_factory](https://github.com/pepoluan/aiosmtpd/tree/catch_factory). Am in process of makening ensuration* it passes all tests on the 7 environments I have.

\* Totally madened up words  (😁)

pepoluan:
## What do these changes do?

Implement mechanism to catch Exceptions raised by `factory()`.

User must be aware that it is now possible for `Controller.start()` to raise Exception caused by _other_ than host:port conflict

## Related issue number

Closes #212 

## Checklist

  - [x] Windows 10 (via PyCharm): `(ALL)`
  - [x] WIndows 10 (via Cygwin): `qa,py36-{nocov,cov}`


pepoluan:
> This pull request **introduces 1 alert** when merging [df48a5a](https://github.com/aio-libs/aiosmtpd/commit/df48a5ace08bbf2e98bcbb8ce93e498bd7fddfd1) into [3b10324](https://github.com/aio-libs/aiosmtpd/commit/3b1032490ade6cb8541a635a641c0e2597970821) - [view on LGTM.com](https://lgtm.com/projects/g/aio-libs/aiosmtpd/rev/pr-af4174d4e8e92d741f5bd2e657e23ec3a6f8715b)
> 
> **new alerts:**
> 
>     * 1 for Default version of SSL/TLS may be insecure

Ugh, it was just for internal check -_-

Okay, let's figure out how to fix this.

pepoluan:
Okay, w.r.t. [this LGTM.com alert](https://lgtm.com/projects/g/aio-libs/aiosmtpd/snapshot/2703f64b7f4f3397af4bccccbf81a67bb63e198a/files/aiosmtpd/controller.py?sort=name&dir=ASC&mode=heatmap#xc6f4a1a0f91be47a:1)

This is difficult.

The [linked LGTM.com documentation for that alert](https://lgtm.com/rules/1507225275976/) specifies that, and I quote:

> Both of the cases above should be updated to use a secure protocol instead, for instance by specifying ssl_version=PROTOCOL_TLSv1_1 as a keyword argument.

The Big Problem™ is that the [Python documentation for `ssl.PROTOCOL_*`](https://docs.python.org/3.6/library/ssl.html#ssl.PROTOCOL_TLSv1_1), stated quite clearly, and I quote:

> Deprecated since version 3.6: OpenSSL has deprecated all version specific protocols. Use the default protocol PROTOCOL_TLS with flags like OP_NO_SSLv3 instead.

PLUS the [Python documentation for `ssl.SSLContext`](https://docs.python.org/3.6/library/ssl.html#ssl.SSLContext) clearly says, and I partially quote:

> Changed in version 3.6: The context is created with secure default values. The options ... OP_NO_SSLv2 (except for PROTOCOL_SSLv2), and OP_NO_SSLv3 (except for PROTOCOL_SSLv3) are set by default.

So it's a clear case that _that_ particular rule of LGTM.com is not up-to-date especially w.r.t. Python 3.6, which we had specified as the minimum supported Python version.

I'm trying to figure out how to continue further on this PR...

pepoluan:
Ugh, the "intermittent coverage fail" again ... let me cherry-pick from PR #211 to fix this ...

.coveragerc
```diff
@@ -19,3 +19,4 @@ exclude_lines =
     pragma: nocover
     pragma: no${PLATFORM}
     pragma: skipif_${PY_39}_py39
+    pragma: skipif_${PY_36}_py36
```
pepoluan, line 22: Needed to 'invert' the steps of getting `all_tasks()`; see [this line](https://github.com/aio-libs/aiosmtpd/pull/213/files#diff-89c287d7da41aad2a025a035e9dddd498ad0ff054ec415bdeac66629ac067b41R146)

If the steps are not inverted, Python 3.7 complains loudly.

tox.ini
```diff
@@ -1,27 +1,25 @@
 [tox]
-envlist = {py36,py37,py38,py39}-{cov,nocov,diffcov}, qa, docs
+envlist = qa, docs, py{36,37,38,39}-{nocov,cov,diffcov}
```
pepoluan, line 2: Faster tests should run earlier so we can quickly press Ctrl-C if we see `qa` or `docs` fail.

tox.ini
```diff
@@ -1,27 +1,25 @@
...
 skip_missing_interpreters = True
 
 [testenv]
 commands =
-    python -c "import sys; print(sys.version); print(sys.executable)"
-    python -c "import os,pprint; os.makedirs('_dynamic', exist_ok=True); pprint.pprint(dict(os.environ), stream=open('_dynamic/ENV."{envname}"', 'wt'))"
+    python housekeep.py setup
```
pepoluan, line 7: Now that I've checked out `housekeep.py` from #202 , let's use that to simplify `tox.ini`

tox.ini
```diff
@@ -1,27 +1,25 @@
...
     nocov: python -m nose2 -v {posargs}
     cov: python -m coverage run -m nose2
     cov: python -m coverage combine
     cov: python -m coverage html --directory=htmlcov/{envname}
     cov: python -m coverage report -m --fail-under=100
-    diffcov: python -m coverage xml
+    cov: python -m coverage xml
```
pepoluan, line 13: XML generation should be part of the `cov` step. This simplifies deps, making full tox faster (and using less disk space)

tox.ini
```diff
@@ -1,27 +1,25 @@
...
-    profile: python -c "open('_dynamic/cPROFILE."{envname}"', 'wb').write(open('_dynamic/cPROFILE', 'rb').read())"
-    python -c "import sys; print(sys.version); print(sys.executable)"
+    python housekeep.py cleanup
 #sitepackages = True
 usedevelop = True
 deps =
-    nose2
-    {cov,diffcov}: coverage
+    !diffcov: nose2
+    cov: coverage
```
pepoluan, line 22: Since `diffcov` now only process XML from `cov`, `diffcov` no longer need `coverage`

tox.ini
```diff
@@ -1,27 +1,25 @@
...
     profile: python -m nose2 --profile {posargs}
-    profile: python -c "open('_dynamic/cPROFILE."{envname}"', 'wb').write(open('_dynamic/cPROFILE', 'rb').read())"
-    python -c "import sys; print(sys.version); print(sys.executable)"
+    python housekeep.py cleanup
 #sitepackages = True
 usedevelop = True
 deps =
-    nose2
-    {cov,diffcov}: coverage
+    !diffcov: nose2
```
pepoluan, line 21: `nose2` is never needed by `diffcov`, let's take it out.

tox.ini
```diff
@@ -33,11 +31,16 @@ setenv =
     py39: INTERP=py39
     pypy3: INTERP=pypy3
     nocov: PYTHONASYNCIODEBUG=1
-    !py39: PY_39=lt
+    py36: PY_36=eq
+    !py36: PY_36=gt
```
pepoluan, line 35: This pair of `setenv`s are required for coverage exclusions on Python>3.6

tox.ini
```diff
@@ -33,11 +31,16 @@ setenv =
...
     py39: PY_39=ge
+    !py39: PY_39=lt
     PLATFORM={env:PLATFORM:linux}
 passenv =
     PYTHON*
+    TRAVIS
+    CI
+    GITHUB*
```
pepoluan, line 43: `TRAVIS` (if set) and `CI` required by `housekeep.py`.

`GITHUB*` added for future needs.

aiosmtpd/tests/test_starttls.py
```diff
@@ -67,6 +67,53 @@ def handle_STARTTLS(self, server, session, envelope):
         return False
 
 
+class EOFingHandler:
```
pepoluan, line 70: This whole class, checked out from #211 (which itself adapted from #202) fixes the "intermittent coverage fail" problem.

aiosmtpd/tests/test_server.py
```diff
@@ -56,3 +58,89 @@ def test_server_attribute(self):
         finally:
             controller.stop()
             self.assertIsNone(controller.server)
+
+
+class TestFactory(unittest.TestCase):
```
pepoluan, line 63: This class contains the tests related to the new `Controller` behavior.

aiosmtpd/smtp.py
```diff
@@ -243,15 +243,10 @@ def connection_lost(self, error):
...
-        if self.session.ssl is not None:            # pragma: nomswin
+        if self.session.ssl is not None:
             # If STARTTLS was issued, return False, because True has no effect
             # on an SSL transport and raises a warning. Our superclass has no
             # way of knowing we switched to SSL so it might return True.
-            #
-            # This entire method seems not to be called during any of the
-            # starttls tests on Windows.  I don't really know why, but it
-            # causes these lines to fail coverage, hence the `nomswin` pragma
-            # above.
```
pepoluan, line 254: All changes in `smtp.py` is related to solving the "intermittent coverage fail" problem.

aiosmtpd/controller.py
```diff
@@ -67,18 +125,33 @@ def start(self):
...
+            raise RuntimeError("Unknown Error, failed to init SMTP server")
 
     def _stop(self):
         self.loop.stop()
         try:
-            _all_tasks = asyncio.Task.all_tasks
-        except AttributeError:   # pragma: skipif_lt_py39
             _all_tasks = asyncio.all_tasks
+        except AttributeError:  # pragma: skipif_gt_py36
+            _all_tasks = asyncio.Task.all_tasks
```
pepoluan, line 149: Try `asyncio.all_tasks` first or else Python 3.7 will complain.

lgtm.yml
```diff
@@ -0,0 +1,9 @@
+queries:
+- exclude: py/insecure-default-protocol
+  # For Python>=3.6, ssl.SSLContext by default uses secure settings.
+  # See https://docs.python.org/3.6/library/ssl.html#ssl.SSLContext
```
pepoluan, line 4: This is a workaround to disable LGTM.com _deprecated_ checking of SSL insecure default protocol.

Sadly, this will only become effective after merging.

pepoluan:
I give up. Can someone **disable LGTM analysis of JavaScript** for aiosmtpd?

lgtm.yml
```diff
@@ -0,0 +1,18 @@
...
+        - ".git"
+        - ".github"
+        - "aiosmtpd"
+        - "build"
+        - "examples"
+        - "htmlcov"
+      filters:
+        - exclude: "**/*.js"
+        - exclude: "**/*.css"
+        - exclude: "**/*.html"
```
pepoluan, line 18: Meant to disable LGTM checking of JavaScript. Either doesn't work, or needs to be merged first to be effective.

pepoluan (commented):
Added comments for explaining bits and pieces.

pepoluan:
I'll test these changes on my 7 test systems, and if they all passes, I'll undraft this.

aiosmtpd/docs/controller.rst
```diff
@@ -242,8 +242,7 @@ Controller API
       class being controlled.  By default, this creates an ``SMTP`` instance,
       passing in your handler and setting the ``enable_SMTPUTF8`` flag.
       Examples of why you would want to override this method include creating
-      an ``LMTP`` server instance instead, or passing in a different set of
-      arguments to the ``SMTP`` constructor.
+      an ``LMTP`` server instance instead, or using your own ``SMTP`` class.
```
pepoluan, line 245: The `server_kwargs` parameter means the last use case for custom Controller is no longer accurate.

aiosmtpd/tests/test_handlers.py
```diff
@@ -17,6 +17,19 @@
...
+def setUpModule():
+    # Needed especially on FreeBSD because socket.getfqdn() is slow on that OS,
+    # and oftentimes (not always, though) leads to Error
+    ModuleResources.enter_context(patch("socket.getfqdn", return_value="localhost"))
+
+
+def tearDownModule():
+    ModuleResources.close()
+
+
```
pepoluan, line 32: Required to stabilize testing on FreeBSD.

`unittest` seems to be _much_ slower than `pytest` on FreeBSD, don't know why.

aiosmtpd/tests/test_server.py
```diff
@@ -56,3 +73,143 @@ def test_server_attribute(self):
...
+            cls.SRProto.__del__ = silencer(cls.old_srp_del)
+        except AttributeError:
+            # Sometimes the __del__ method is ... missing?? What???
+            pass
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        # gc.collect() hinted in https://stackoverflow.com/a/25067818/149900
+        # Probably to remove leftover "Exception ignored"?
+        gc.collect()
```
pepoluan, line 130: Without `gc.collect()` somehow the "Exception ignored" bit survived capture in `silencer()`. Not sure why.

pepoluan (commented):
Additional comments on particularly interesting bits.

pepoluan:
I hope this can be merged soon... I have some PR-to-be (i.e., still on my laptop/repo and not turned into a PR), that kind of relies on this PR...

aiosmtpd/docs/controller.rst
```diff
@@ -230,7 +232,24 @@ Controller API
...
       raised if the server does not start within the *ready_timeout*, or if
-      any other exception occurs in while creating the server.
+      any other exception occurs in :meth:`factory` while creating the server.
+
+      .. important::
+
+         If :meth:`start` raises an Exception,
+         :class:`Controller` does not automatically perform cleanup,
+         to support deep inspection post-exception (if you wish to do so.)
+         Cleanup must still be performed manually by calling :meth:`stop`
```
pepoluan, line 242: "Semantic Linefeeds", as suggested (and somewhat evangelized) by @brandon-rhodes , explained very well in [his blog post](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

pepoluan (commented):
Added some comments. The "Semantic Linefeeds" idea is great; let's have more of that!

pepoluan:
This has been passing checks and waiting review for almost 2 weeks; I'm merging within 24h.

PR #218, opened 2020-12-14
Improved authenticator system

### LINKED ISSUE #219: Implement support for "AUTH LOGIN <username>"
Although the [documentation for AUTH LOGIN](https://www.ietf.org/archive/id/draft-murchison-sasl-login-00.txt) implies that `AUTH LOGIN` should not have args following it, Python's SMTP client implementation in `smtplib.SMTP` assumes so.

As a result, the strict process in current `master`'s AUTH LOGIN implementation will cause failure if `smtplib.SMTP` tries to auth using the LOGIN mechanism.

To explain visually:

This is how the transaction **_should_** be:

```
C: AUTH LOGIN
S: 334 VXNlciBOYW1lAA==
C: <base64-encoded username>
S: 334 UGFzc3dvcmQA
C: <base64-encoded password>
```

This is how `smtplib.SMTP` _wants_ it to be (when `initial_response_ok` is set to `True`, which is the default):

```
C: AUTH LOGIN <base-64-encoded username>
S: 334 UGFzc3dvcmQA
C: <base64-encoded password>
```

But, if `initial_response_ok=False` is used, then `smtplib.SMTP` will do it like this:

```
C: AUTH LOGIN
S: 334 VXNlciBOYW1lAA==
C: <base64-encoded password>  # yes, "password" not "username"
S: 334 UGFzc3dvcmQA
C: <base64-encoded password>
```

This is discussed in [**bpo-27820**](https://bugs.python.org/issue27820) (to which I've suggested a fix)

pepoluan:
## What do these changes do?

In previous implementation of `auth_callback`, the authenticator callback function is only able to see three information: Mechanism of authentication, login/username, and password. The function cannot determine additional information such as the peer IP address, state of TLS, etc.

The proposed changes in this PR adds, as an alternative to `auth_callback`, a new parameter named `authenticator`.

The function passed to `authenticator` will not only see the three information as seen by auth_callback, it will also see server, session, and envelope (similar to how handler hooks are called -- although in AUTH case, envelope is likely empty and useless).

In addition, the return value of `authenticator` function is more informative instead of the overloaded, non-intuitive `_TriStateType`.

**In addition,** this PR also now supports `AUTH LOGIN <b64encoded-username>` because Python's `smtplib.SMTP` class has a buggy implementation of `AUTH LOGIN` (see [**bpo-27820**](https://bugs.python.org/issue27820)).

Should be None. `authenticator` does not replace `auth_callback`; current users relying on `auth_callback` can continue doing so, while future users can use the much more powerful `authenticator` system.

## Related issue number

Closes #219

## Checklist


pepoluan:
I'm keeping this as Draft until I have the time to properly test the changes on my test systems.

pepoluan:
> 
> 
> I'm keeping this as Draft until I have the time to properly test the changes on my test systems.

Sorry, had been a very intense sprint in my job for the past 2.5 weeks. I can now pick up on things.

Testing progress:
- [ ] Windows 10 (via PyCharm tox runner): `(ALL)`

(This post is undergoing editing)

.github/workflows/unit-testing-and-coverage.yml
```diff
@@ -25,15 +25,22 @@ jobs:
...
+            #     someone apparently defined the term "stable" as "still receiving
+            #     bugfixes" so I had to change to prevent confusion.
+            #
+            # Also, Windows users can change their Python anytime since it's not part of the OS.
+            # Hence there is no benefit in testing all released Python versions.
+            #
+            # In addition, pypy3 implementation in Windows is ... complicated. We should stay
+            # away from all those complications; users choosing to use pypy3 on Windows to run
+            # aiosmtpd should be considered advanced, know what they're doing, and ready for
+            # all the consequences
```
pepoluan, line 43: A looong explanation on why we use 3.7 on Windows ... a self-made term "frozen".

aiosmtpd/docs/smtp.rst
```diff
@@ -103,12 +103,14 @@ override to provide additional responses.
...
 ========
 
-.. class:: SMTP(handler, *, data_size_limit=33554432, enable_SMTPUTF8=False, \
-   decode_data=False, hostname=None, ident=None, tls_context=None, \
-   require_starttls=False, timeout=300, auth_required=False, \
-   auth_require_tls=True, auth_exclude_mechanism=None, auth_callback=None, \
+.. class:: SMTP(handler, *, data_size_limit=33554432, enable_SMTPUTF8=False, decode_data=False, \
+   hostname=None, ident=None, tls_context=None, require_starttls=False, timeout=300, \
+   auth_required=False, auth_require_tls=True, auth_exclude_mechanism=None, \
+   auth_callback=None, authenticator=None, \
```
pepoluan, line 109: Split overly-long signature line.

aiosmtpd/docs/smtp.rst
```diff
@@ -4,10 +4,10 @@
  The SMTP class
 ================
 
-At the heart of this module is the ``SMTP`` class.  This class implements the
-`RFC 5321 <http://www.faqs.org/rfcs/rfc5321.html>`_ Simple Mail Transport
-Protocol.  Often you won't run an ``SMTP`` instance directly, but instead will
-use a :ref:`controller <controller>` instance to run the server in a subthread.
+At the heart of this module is the ``SMTP`` class.
+This class implements the :rfc:`5321` Simple Mail Transport Protocol.
```
pepoluan, line 8: Let's use the `:rfc:` directive more often!

aiosmtpd/docs/smtp.rst
```diff
@@ -302,7 +332,4 @@ advertised, and the ``STARTTLS`` command will not be accepted.
 ``False``.
 
 .. _StreamReaderProtocol: https://docs.python.org/3/library/asyncio-stream.html#streamreaderprotocol
-.. _`RFC 3207`: http://www.faqs.org/rfcs/rfc3207.html
-.. _`RFC 2821`: https://www.ietf.org/rfc/rfc2821.txt
```
pepoluan, line 306: The `:rfc:` directive takes care of formatting and linking.

aiosmtpd/docs/smtp.rst
```diff
@@ -302,7 +332,4 @@ advertised, and the ``STARTTLS`` command will not be accepted.
...
 .. _`asyncio transport`: https://docs.python.org/3/library/asyncio-protocol.html#asyncio-transport
-.. _ssl: https://docs.python.org/3/library/ssl.html
```
pepoluan, line 308: The `:mod:` directive takes care of formatting and linking.

aiosmtpd/smtp.py
```diff
@@ -550,50 +582,71 @@ async def smtp_AUTH(self, arg: str) -> None:
...
-            await self.push("538 5.7.11 Encryption required for requested "
-                            "authentication mechanism")
+            return await self.push("538 5.7.11 Encryption required for requested "
+                                   "authentication mechanism")
         elif self.authenticated:
-            await self.push('503 Already authenticated')
+            return await self.push('503 Already authenticated')
         elif not arg:
-            await self.push('501 Not enough value')
-        else:
```
pepoluan, line 561: Code is pushed to the right and honestly looks really ugly. Especially as it has gained complexity due to having to cater to the new `authenticator` system while maintaining backward compatibility with the old `auth_callback` system.

So I changed all the naked `await`s up there with `return await`

aiosmtpd/smtp.py
```diff
@@ -615,78 +668,93 @@ async def _auth_interact(self, server_message) -> _TriStateType:
...
+        return self._authenticate("PLAIN", _LoginPassword(login, password))
 
-    async def auth_LOGIN(self, _, args: List[str]):
-        login: _TriStateType
-        # 'User Name\x00'
-        login = await self._auth_interact("334 VXNlciBOYW1lAA==")
+    async def auth_LOGIN(self, _, args: List[str]) -> AuthResult:
+        if len(args) > 1:
+            try:
+                login = b64decode(args[1].encode(), validate=True)
```
pepoluan, line 740: This piece of code handles the -- actually RFC non-compliant -- `AUTH LOGIN <b64_username>` procedure.

Well, actually there is _no_ RFC for `AUTH LOGIN` ... only [a draft RFC](https://www.ietf.org/archive/id/draft-murchison-sasl-login-00.txt) that never gets adopted/promoted into full RFC status. So in this case, we can be more lenient with implementation of `AUTH LOGIN`.

(Also, Python>=3.5 `smtplib.SMTP` is actually buggy; see [**bpo-27820**](https://bugs.python.org/issue27820))

aiosmtpd/smtp.py
```diff
@@ -550,50 +582,71 @@ async def smtp_AUTH(self, arg: str) -> None:
...
+                log.debug(f"Using builtin auth_ hook for {mechanism}")
+            else:
+                log.debug(f"Using handler auth_ hook for {mechanism}")
+            # Pass 'self' to method so external methods can leverage this
+            # class's helper methods such as push()
+            auth_result = await auth_method.method(self, args)
+            log.debug(f"auth_{mechanism} returned {auth_result}")
+
+            # New system using `authenticator` and AuthResult
+            if isinstance(auth_result, AuthResult):
```
pepoluan, line 617: This test handles backward- and forward-compatibility for `auth_*` methods.

aiosmtpd/tests/test_smtp.py
```diff
@@ -969,12 +1042,55 @@ def test_ehlo(self):
...
+            client.password = "goodpass"
+            client.ehlo("example.com")
+            client.auth("login", client.auth_login)
+        self.assertEqual(b"gooduser", auth_peeker.login)
+        self.assertEqual(b"goodpass", auth_peeker.password)
+        self.assertEqual("LOGIN", auth_peeker.mechanism)
+
+    # Mark this as expectedFailure because smtplib.SMTP implementation in Python>=3.5
+    # is buggy. See bpo-27820
+    @unittest.expectedFailure
```
pepoluan, line 1083: Until [**bpo-27820**](https://bugs.python.org/issue27820) is fixed, keep this `expectedFailure` line. When the bug is fixed, we'll ... figure something out.

pepoluan (commented):
I've left comment on possibly interesting points.

aiosmtpd/smtp.py
```diff
@@ -119,23 +147,27 @@ class SMTP(asyncio.StreamReaderProtocol):
...
-            auth_required=False,
-            auth_require_tls=True,
-            auth_exclude_mechanism: Optional[Iterable[str]] = None,
-            auth_callback: Callable[[str, bytes, bytes], bool] = None,
-            loop=None,
-    ):
+    # base64-encoded 'User Name\x00'
+    AuthLoginUsernameChallenge = "VXNlciBOYW1lAA=="
+    # base64-encoded 'Password\x00'
+    AuthLoginPasswordChallenge = "UGFzc3dvcmQA"
```
pepoluan, line 153: Some noncompliant SMTP client implementation expects "Username:" and "Password:". To cater for those noncompliant clients, the AUTH LOGIN challenges are customizable.

(Standard says that clients should totally ignore the challenges.)

aiosmtpd/tests/test_smtp.py
```diff
@@ -58,37 +69,70 @@ def factory(self):
...
     async def auth_DONT(
             self, server, args
     ):
         return MISSING
 
 
 class PeekerAuth:
-    def __init__(self):
-        self.login = None
-        self.password = None
```
pepoluan, line 83: No need for __init__ if all it does is initializing some instance attributes.

Just convert those instance attributes into class attributes; first time the instance attribute gets assigned to, an instance attribute will be created from the class attribute.

pepoluan (commented):
I've left comments on likely interesting points.

aiosmtpd/docs/handlers.rst
```diff
@@ -183,7 +183,7 @@ those methods of the handler instance will override the built-in methods.
 
   :boldital:`server` is the instance of the ``SMTP`` class invoking the AUTH hook.
   This allows the AUTH hook implementation to invoke facilities such as the
-  ``push()`` and ``_auth_interact()`` methods.
+  ``push()`` and ``challenge_auth()`` methods.
```
pepoluan, line 186: Renamed the function so it doesn't start with "_"

Had to be creative here because "auth_*" prefix has its own meaning...

aiosmtpd/smtp.py
```diff
@@ -1,5 +1,6 @@
 import re
 import ssl
+import attr
```
pepoluan, line 3: Using `attrs` to enforce some behaviors.

aiosmtpd/testing/helpers.py
```diff
@@ -65,19 +65,25 @@ def start(plugin):
...
 
 
-def assert_auth_success(testcase: TestCase, code, response):
-    testcase.assertEqual(code, 235)
-    testcase.assertEqual(response, b"2.7.0 Authentication successful")
+def assert_auth_success(testcase: TestCase, *response):
+    testcase.assertEqual(
+        (235, b"2.7.0 Authentication successful"),
+        response
+    )
```
pepoluan, line 72: Compare tuple against tuple directly so helpful error messages won't get hidden.

Also for `assert_auth_invalid` and `assert_auth_required` below.

aiosmtpd/tests/test_smtp.py
```diff
@@ -48,7 +57,7 @@ def tearDownModule():
     ModuleResources.close()
 
 
-def authenticator(mechanism, login, password):
+def auth_callback(mechanism, login, password):
```
pepoluan, line 60: Rename to prevent confusion with the brand new `authenticator` system.

aiosmtpd/tests/test_smtp.py
```diff
@@ -984,29 +1064,75 @@ def test_ehlo(self):
...
+                (
+                    b'AUTH DENYFALSE DENYMISSING LOGIN NONE NULL PLAIN '
+                    b'WITH-DASH WITH-MULTI-DASH WITH_UNDERSCORE'
+                ),
                 b'HELP',
-            )
-            for actual, expected in zip(lines, expecteds):
-                self.assertEqual(actual, expected)
+            ]
+            assert lines == expecteds
```
pepoluan, line 1077: Again, do direct compare list against list. The line-by-line comparison sometimes resulted in confusing failures. By comparing whole list against whole list, I can get a better grasp of what's going on.

Also, transition to pytest's `assert`-based testing.

aiosmtpd/tests/test_smtp.py
```diff
@@ -1019,11 +1145,11 @@ def test_auth_login_null_credential(self):
...
-            self.assertEqual(auth_peeker.login, None)
-            self.assertEqual(auth_peeker.password, None)
-            code, response = client.mail("alice@example.com")
-            sess: SMTPSess = self.handler.session
-            self.assertEqual(sess.login_data, b"")
+            assert auth_peeker.mechanism == "LOGIN"
+            assert auth_peeker.login == b""
+            assert auth_peeker.password == b""
+            response = client.mail("alice@example.com")
+            assert response == (250, b"OK")
```
pepoluan, line 1152: Replace the pointless "let's check that login_data doesn't change" test with a test that actually tests that 'protected' commands are accepted.

Also, migrated some tests to pytest's `assert` methodology.

aiosmtpd/tests/test_smtp.py
```diff
@@ -1154,9 +1344,8 @@ def test_help_authenticated(self):
                 b64encode(b'\0goodlogin\0goodpasswd').decode()
             )
             assert_auth_success(self, code, response)
-            code, response = client.docmd('HELP')
-            self.assertEqual(code, 250)
-            self.assertEqual(response, SUPPORTED_COMMANDS_NOTLS)
```
pepoluan, line 1159: Again, change from piece-by-piece comparison to direct tuple-vs-tuple comparison, in order to not hide possibly helpful error messages.

Also, migrate to pytest's `assert` syntax.

pepoluan:
Huh, what's up with MacOS runner in GA? It's now intermittently failing. First time it failed with Pyhon 3.8, afterwards it failed with Python 3.6.

Luckily, with GA I can just ask it to "re-run all jobs" and 3rd time _was_ the charm.

But this makes me wonder if I should drop the MacOS runner in the GA yml ...

pepoluan:
As usual, I'm marinating this for 2x24 before squamerging.

PR #222, opened 2020-12-23
Command Call Limit

### LINKED ISSUE #145: Misbehaving clients are never disconnected

pepoluan:
## What do these changes do?

Implements an **optional** mechanism to limit the number of _same_ command calls.

* Can specify a "global" limit or differing per-command limit
* Maintains backward compatibility for users expecting no command call limitation

Nothing for existing users. Backward compatibility is maintained.

## Related issue number

Closes #145 

## Checklist


aiosmtpd/smtp.py
```diff
@@ -403,6 +433,17 @@ async def _handle_client(self):
                     # RFC3207 part 4
                     await self.push('530 Must issue a STARTTLS command first')
                     continue
+
+                if self._enforce_call_limit:
+                    budget = call_limit[command]
+                    if budget < 1:
+                        await self.push(
+                            f"503 {command} sent too many times"
```
TimWolla, line 441: I'd say that an error 421 would be more fitting here: https://tools.ietf.org/html/rfc5321#section-3.8

>    An SMTP server MUST NOT intentionally close the connection under
>   normal operational circumstances (see Section 7.8) except:
>   […]
>    o  After detecting the need to shut down the SMTP service and
>      returning a 421 response code.  This response code can be issued
>      after the server receives any command or, if necessary,
>      asynchronously from command receipt (on the assumption that the
>      client will receive it after the next command is issued).

TimWolla:
This should probably also handle the number of erroneous commands, e.g. repeatedly sending `BOGUS`.

aiosmtpd/smtp.py
pepoluan: That is only "under normal operational circumstances".

I'm using `503` because `5yz` is the recommended status code _during an attack_: https://tools.ietf.org/html/rfc5321#section-7.8

TimWolla: I smell a footgun here. Consider a user of aiosmtpd putting `RCPT` with a limit of 10 into the configuration. I mean who needs more than 10 recipients for an email?

What they might miss is the fact that a single TCP connection allows to deliver multiple distinct emails and clients will do this. Consider a higher traffic ESP with multiple recipients at the destination MTA. Now that ESP sends 11 emails with a single recipient, receiving a 503 in response and as a result permanently drops the email for the 11th recipient.

Using a 421 is the safe option here:

- Actual attackers do not care about the status codes, they'll just try over and over again.
- Legitimate mailers usually behave properly, never triggering the limit.
- In edge cases (such as my RCPT example) aiosmtpd will fail without causing lost mail.
- Additionally the additional load caused by a legitimate client retrying something that will never succeed (e.g. setting `DATA` to `0`) will be negligible.

Also I consider "What would Postfix do" a good test to verify the implemented behavior makes sense:

- Postfix increments the error counter for each "useless" command after a certain limit.
- When the error counter reaches a specific number it drops the connection after sending a 421.
- The error counter is reset after each email that was delivered successfully.

pepoluan: Hmm, makes sense.

Alright, I'll change that to `421 4.7.0` then.

> The error counter is reset after each email that was delivered successfully.

I don't think this one can be implemented in aiosmtpd though, especially since delivery success (or not) totally depends on user-supplied `handle_DATA`.

`aiosmtpd`, as it is now, is a "half-MTA/MSA", providing the API to receive email, but not having mechanism to actually forward the email to the next MTA / MDA.

pepoluan:
> 
> 
> This should probably also handle the number of erroneous commands, e.g. repeatedly sending `BOGUS`.

It's already built-in.

The "budget check" happens before the method search-and-invoke, so even bogus commands will be limited.

pepoluan:
Doing testing for latest changes + merges on my test systems:

- [ ] Windows 10 (via PSCore 7.1.0): `(ALL)`

**_This comment is undergoing editing_**

aiosmtpd/smtp.py
```diff
@@ -59,11 +59,13 @@ class _DataState(enum.Enum):
     "AuthMechanismType",
     "MISSING",
 ]  # Will be added to by @public
-__version__ = '1.2.3a4'
+__version__ = '1.3.0a1'
 __ident__ = 'Python SMTP {}'.format(__version__)
 log = logging.getLogger('mail.log')
 
 
+BOGUS_LIMIT = 5
```
pepoluan, line 67: Yes, "bogus" commands are hard-limited to max of 5 per session.

I don't see why we need to make this customizable.

aiosmtpd/smtp.py
```diff
@@ -451,8 +484,31 @@ async def _handle_client(self):
...
+                    call_limit[command] = budget - 1
+
                 method = self._smtp_methods.get(command)
                 if method is None:
+                    log.warning("%r unrecognised: %s", self.session.peer, command)
+                    bogus_budget -= 1
+                    if bogus_budget < 1:
+                        log.warning("%r too many bogus commands", self.session.peer)
+                        await self.push(
+                            "502 5.5.1 Too many unrecognized commands, goodbye."
```
pepoluan, line 508: The status code is the same as what Gmail uses.

pepoluan:
If no objections within the next 2x24, I'll squamerge this PR.

PR #225, opened 2020-12-27
Workaround to specify additional AUTH mechanisms with dash

### LINKED ISSUE #224: Current AUTH hook implementation cannot support mechanisms with dash "-" in the name
The current implementation simply searches for `auth_*` methods in the handler, chop off the `auth_` part, and present the chopped name as is to the list of supported authentications.

This means that AUTH mechanisms such as "CRAM-MD5" and "DIGEST-MD5" are not representable.

Fixing this is a bit tricky; the simple trick of replacing "_" with "-" will result in SASL mechanisms that _does_ contain underscore in their names (see [SASL mechanisms registry on IANA](https://www.iana.org/assignments/sasl-mechanisms/sasl-mechanisms.xhtml)) to be not representable.

My suggestion would be to add another decorator (similar to the existing `@syntax` decorator) that will add an attribute, say `__auth_mechanism_name__` ; if the attribute exists, use that. If the attribute does not exist, use the existing name-mangling strategy.

pepoluan:
## What do these changes do?

The current implementation for AUTH Hooks does not allow AUTH mechanisms with dash(es) in their names.

For example, it is not possible to create a handler method for `CRAM-MD5` or `DIGEST-MD5`, because the code simply takes the method name beginning with "auth_" and chop off that part.

This PR provides two methods of specifying dashes:

  * Using "double underscore", so `MECH__NAME__WITH__DASH` ==> `MECH-NAME-WITH-DASH`
  *  Using `@auth_mechanism(actual_name)` decorator

None for existing users unless they had subclassed `SMTP` to implement this feature.

## Related issue number

Closes #224 

## Checklist

  - [x] Windows 10 (via PyCharm runner): `(ALL)`
  - [x] Ubuntu 18.04 on WSL 1.0: `(ALL)` + `pypy3-{nocov,cov}`
  - [x] Ubuntu 18.04 on VBox: `(ALL)` + `pypy3-{nocov,cov}`
  - [x] Ubuntu 20.04 on VBox: `(ALL)` + `pypy3-{nocov,cov}`


pepoluan:
Will undraft after I do testing on my test systems.

aiosmtpd/smtp.py
```diff
@@ -61,6 +65,7 @@ class _DataState(enum.Enum):
 EMPTYBYTES = b''
 MISSING = _Missing()
 NEWLINE = '\n'
+VALID_AUTHMECH = re.compile(r"[A-Z0-9_-]+\Z")
```
pepoluan, line 68: We can revisit this as a bugfix if there is actually a SASL mechanism which has characters outside of these.

pepoluan (commented):
Left some comments.

.gitignore
```diff
@@ -30,3 +30,4 @@ _dynamic/
 coverage.xml
 diffcov.html
 .python-version
+.pytype/
```
pepoluan, line 33: Workdir used by `pytype`

aiosmtpd/smtp.py
```diff
@@ -179,21 +224,25 @@ def __init__(
...
         self._auth_required = auth_required
-        self.authenticated = False
         # Get hooks & methods to significantly speedup getattr's
         self._auth_methods: Dict[str, _AuthMechAttr] = {
-            m.replace("auth_", ""): _AuthMechAttr(getattr(h, m), h is self)
-            for h in (self, handler)
-            for m in dir(h)
-            if m.startswith("auth_")
+            getattr(
+                mfunc, "__auth_mechanism_name__",
```
pepoluan, line 230: Option 1: Decorated auth hooks

aiosmtpd/smtp.py
```diff
@@ -179,21 +224,25 @@ def __init__(
...
+                mname.replace("auth_", "").replace("__", "-")
```
pepoluan, line 231: Option 2: Double-underscored auth hooks

waynew (approved):
🚀

pepoluan:
I'm gonna let this marinate for 2x24 before squamerging, just in case 😊

pepoluan:
First squamerge of 2021, yay! 😁

PR #232, opened 2021-01-15
BUGFIX: RFC4954 behavior for lone "="

pepoluan:
## What do these changes do?

When implementing AUTH support, I misunderstood the meaning of `=` (lone equal sign) in [RFC 4954](https://tools.ietf.org/html/rfc4954).

This has been fixed to strictly comply with RFC 4954 and also [RFC 4616](https://tools.ietf.org/html/rfc4616)

This mostly impacts `AUTH PLAIN` which requires a special null-separated struct.

If user (of this library, not an SMTP client) wants to test AUTH PLAIN with null login, then sending `AUTH PLAIN =` or as a response to the challenge will no longer work. This shouldn't be a problem if test is done using Standard Lib's `smtplib.SMTP`, because `smtplib.SMTP.auth_plain` already properly creates the necessary null-separated struct.

## Related issue number

None

## Checklist

### DIFF
--- a/aiosmtpd/docs/handlers.rst
+++ b/aiosmtpd/docs/handlers.rst
@@ -183,7 +183,7 @@ those methods of the handler instance will override the built-in methods.
 
   :boldital:`server` is the instance of the ``SMTP`` class invoking the AUTH hook.
   This allows the AUTH hook implementation to invoke facilities such as the
-  ``push()`` and ``_auth_interact()`` methods.
+  ``push()`` and ``challenge_auth()`` methods.
 
   :boldital:`args` is a list of string split from the string after the ``AUTH`` command.
   ``args[0]`` is always equal to ``MECHANISM``.
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -795,9 +795,6 @@ class SMTP(asyncio.StreamReaderProtocol):
     #      - 'identity' is not always username, depending on the auth mecha-
     #        nism. Might be a session key, a one-time user ID, or any kind of
     #        object, actually.
-    #      - If the client provides "=" for username during interaction, the
-    #        method MUST return b"" (empty bytes) NOT None, because None has been
-    #        used to indicate error/login failure.
     # 3. Auth credentials checking is performed in the auth_* methods because
     #    more advanced auth mechanism might not return login+password pair
     #    (see #2 above)
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -767,7 +767,10 @@ class SMTP(asyncio.StreamReaderProtocol):
 
     _334_PREFIX = re.compile(r"^334 ")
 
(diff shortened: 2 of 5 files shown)


aiosmtpd/docs/handlers.rst
```diff
@@ -183,7 +183,7 @@ those methods of the handler instance will override the built-in methods.
 
   :boldital:`server` is the instance of the ``SMTP`` class invoking the AUTH hook.
   This allows the AUTH hook implementation to invoke facilities such as the
-  ``push()`` and ``_auth_interact()`` methods.
+  ``push()`` and ``challenge_auth()`` methods.
```
pepoluan, line 186: Deprecating `_auth_interact()` because it's 'private'. Couldn't start with `auth_` though, so used a new name.

aiosmtpd/smtp.py
```diff
@@ -383,10 +384,14 @@ def _set_rset_state(self):
...
-    async def push(self, status):
-        response = bytes(
-            status + '\r\n', 'utf-8' if self.enable_SMTPUTF8 else 'ascii')
-        self._writer.write(response)
+    async def push(self, status: AnyStr):
+        if isinstance(status, str):
+            response = bytes(
+                status, 'utf-8' if self.enable_SMTPUTF8 else 'ascii')
+        else:
+            response = status
```
pepoluan, line 392: `SMTP.push()` is now a bit smarter and can handle `str` or `bytes`.

After all, internally it performs a str.encode() so why not enable it to handle bytes directly?

aiosmtpd/smtp.py
```diff
@@ -723,18 +728,34 @@ async def smtp_AUTH(self, arg: str) -> None:
...
+        # See https://tools.ietf.org/html/rfc4954#page-4 ¶ 5
+        await self.push(
+            b"334 " + (b64encode(challenge) if encode_to_b64 else challenge)
+        )
         line = await self._reader.readline()
-        blob = line.strip()
-        # '=' and '*' handling are in accordance with RFC4954
-        if blob == b"=":
-            log.debug("%r responded with '='", self.session.peer)
-            return None
```
pepoluan, line 734: This is my great mistake; I misunderstood RFC4954.

A sole equal sign `=` is supposed to be Base64-decoded _first_, and from there we get the empty data `b""`

aiosmtpd/smtp.py
```diff
@@ -723,18 +728,34 @@ async def smtp_AUTH(self, arg: str) -> None:
             if status is not None:  # pragma: no branch
                 await self.push(status)
 
-    async def _auth_interact(self, server_message) -> _TriStateType:
-        blob: bytes
-        await self.push(server_message)
+    async def challenge_auth(
+            self,
+            challenge: AnyStr,
+            encode_to_b64: bool = True,
```
pepoluan, line 734: Most AUTH challenges are Base64-encoded. This particular feature of `challenge_auth` enables an easier-to-read implementaton of challenges, as can be seen in the `auth_LOGIN` method.

PR #234, opened 2021-01-17
Hide password in logs

### LINKED ISSUE #233: Possible Security Issue: AUTH credentials visible in log
I was fooling around with a branch, and decided to activate logging.

Found out that all `AUTH` commands are logged as-is, without sanitizing. Even if logging level set to `INFO`.

Not a problem if using multistep authentication (i.e., `AUTH mech` then wait for `334 ` then respond; the `_auth_interact` method does not log client response), but a problem if using one-step authentication (e.g., `AUTH PLAIN <encoded_username_and_password>`)

I fear this might cause to inadvertent leak of sensitive information. For example, if log is being forwarded to an external log collector (e.g., CloudWatch Log on AWS, DataDog, etc.)

I will submit an URGENT PR to fix this.

pepoluan:
## What do these changes do?

Sanitize command line before logging.

This prevents leakage of sensitive information.

Only if user parses the log output.

If user _absolutely_ needs to capture the SMTP credentials, user can:

  1. Override the `auth_*` methods and/or set `log_client_response=True` on calls to `SMTP._auth_interact()` (for custom, non-built-in AUTH mechanisms), or
  2. Perform the logging within the `auth_callback` function

## Related issue number

Closes #233 

## Checklist

   - [x] FreeBSD 12.2 [1] on VBox: `(ALL)` + `pypy3-{nocov,cov,diffcov}`

[1] [12.1 is being EOL-ed](https://www.freebsd.org/security/index.html#sup). Side benefit: Since I decided to just wipe my VM and reinstall from scratch, lo and behold! Now all `pypy3` tests work! And the icing on the cake is that the whole tox suite runs much faster than before!

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -76,6 +76,11 @@ VALID_AUTHMECH = re.compile(r"[A-Z0-9_-]+\Z")
 # https://tools.ietf.org/html/rfc3207.html#page-3
 ALLOWED_BEFORE_STARTTLS = {"NOOP", "EHLO", "STARTTLS", "QUIT"}
 
+# Auth hiding regexes
+CLIENT_AUTH_B = re.compile(
+    br"(?P<authm>\s*AUTH\s+\S+[^\S\r\n]+)(\S+)(?P<crlf>(?:\r\n)?)", re.IGNORECASE
+)
+
 # endregion
 
 
@@ -160,6 +165,14 @@ def is_int(o):
 @public
 class TLSSetupException(Exception):
     pass
+
+
+@public
+def sanitize(text: bytes) -> bytes:
+    m = CLIENT_AUTH_B.match(text)
+    if m:
+        return m.group("authm") + b"********" + m.group("crlf")
+    return text
 
 
 @public
@@ -431,10 +444,10 @@ class SMTP(asyncio.StreamReaderProtocol):
                     # send error response and read the next command line.
                     await self.push('500 Command line too long')
                     continue
-                log.debug('_handle_client readline: %r', line)
+                log.debug('_handle_client readline: %r', sanitize(line))
                 # XXX this rstrip may not completely preserve old behavior.
                 line = line.rstrip(b'\r\n')
-                log.info('%r >> %r', self.session.peer, line)
+                log.info('%r >> %r', self.session.peer, sanitize(line))
                 if not line:
                     await self.push('500 Error: bad syntax')
                     continue
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -60,7 +60,7 @@ __all__ = [
     "AuthMechanismType",
     "MISSING",
 ]  # Will be added to by @public
-__version__ = '1.2.3'
+__version__ = '1.2.4a1'
 __ident__ = 'Python SMTP {}'.format(__version__)
 log = logging.getLogger('mail.log')
 
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -30,7 +30,8 @@ from warnings import warn
 # region #### Custom Data Types #######################################################
 
 class _Missing:
-    pass
+    def __repr__(self):
+        return "MISSING"
 
 
 class _AuthMechAttr(NamedTuple):
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -186,6 +186,15 @@ def sanitize(text: bytes) -> bytes:
 
 
 @public
+def sanitized_log(func: Callable, msg: AnyStr, *args, **kwargs):
+    sanitized_args = [
+        sanitize(a) if isinstance(a, bytes) else a
+        for a in args
+    ]
+    func(msg, *sanitized_args, **kwargs)
+
+
+@public
 class SMTP(asyncio.StreamReaderProtocol):
     command_size_limit = 512
     command_size_limits = collections.defaultdict(
@@ -458,10 +467,10 @@ class SMTP(asyncio.StreamReaderProtocol):
                     # send error response and read the next command line.
                     await self.push('500 Command line too long')
                     continue
-                log.debug('_handle_client readline: %r', sanitize(line))
+                sanitized_log(log.debug, '_handle_client readline: %r', line)
                 # XXX this rstrip may not completely preserve old behavior.
                 line = line.rstrip(b'\r\n')
-                log.info('%r >> %r', self.session.peer, sanitize(line))
+                sanitized_log(log.info, '%r >> %r', self.session.peer, line)
                 if not line:
                     await self.push('500 Error: bad syntax')
                     continue
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -79,7 +79,15 @@ ALLOWED_BEFORE_STARTTLS = {"NOOP", "EHLO", "STARTTLS", "QUIT"}
 
 # Auth hiding regexes
 CLIENT_AUTH_B = re.compile(
-    br"(?P<authm>\s*AUTH\s+\S+[^\S\r\n]+)(\S+)(?P<crlf>(?:\r\n)?)", re.IGNORECASE
+    # Matches "AUTH" <mechanism> <whitespace_but_not_\r_nor_\n>
+    br"(?P<authm>\s*AUTH\s+\S+[^\S\r\n]+)"
+    # Param to AUTH <mechanism>. We only need to sanitize if param is given, which
+    # for some mechanisms contain sensitive info. If no param is given, then we
+    # can skip (match fails)
+    br"(\S+)"
+    # Optional bCRLF at end. Why optional? Because we also want to sanitize the
+    # stripped line. If no bCRLF, then this group will be b""
+    br"(?P<crlf>(?:\r\n)?)", re.IGNORECASE
 )
 
 # endregion
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -737,10 +737,26 @@ class SMTP(asyncio.StreamReaderProtocol):
             if status is not None:  # pragma: no branch
                 await self.push(status)
 
-    async def _auth_interact(self, server_message) -> _TriStateType:
+    async def _auth_interact(
+            self,
+            server_message,
+            log_client_response: bool = False
+    ) -> _TriStateType:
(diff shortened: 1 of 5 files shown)


aiosmtpd/smtp.py
```diff
@@ -30,7 +30,8 @@
 # region #### Custom Data Types #######################################################
 
 class _Missing:
-    pass
+    def __repr__(self):
+        return "MISSING"
```
pepoluan, line 34: For better representation in the logs.

aiosmtpd/smtp.py
```diff
@@ -723,10 +737,26 @@ async def smtp_AUTH(self, arg: str) -> None:
...
+            log_client_response: bool = False
+    ) -> _TriStateType:
+        """
+        Sends a challenge, and wait for client response.
+
+        :param server_message: Challenge to send
+        :param log_client_response: Perform logging of client's response.
+            WARNING: Might cause leak of sensitive information! Do not turn on
+            unless _absolutely_ necessary!
+        """
```
pepoluan, line 752: Come to think of it, logging AUTH interaction _might_ be useful in some situations.

So rather than hardcoding no logging, we'll provide a default-disabled logging here.

aiosmtpd/testing/helpers.py
```diff
@@ -65,9 +65,8 @@ def start(plugin):
         warnings.filterwarnings('always', category=ResourceWarning)
 
 
-def assert_auth_success(testcase: TestCase, code, response):
-    testcase.assertEqual(code, 235)
-    testcase.assertEqual(response, b"2.7.0 Authentication successful")
```
pepoluan, line 70: Previous logic stops on mismatch of code. Hiding possibly important info contained in the message.

So I rewrite the comparo to directly compare the tuples; no more hidden info :-)

pepoluan (commented):
Some comments have been left for possibly interesting points.

pepoluan:
I am so happy that I took the 'nuclear' option and just nuked my FreeBSD 12.1 VM instead of upgrading it to 12.2.

Now all testenvs work, including `pypy3-cov` and `pypy3-diffcov`. And the tests run _much_ faster!

I'm not blaming 12.1 for the problems I've experienced; I haven't used FreeBSD for quite some time before I decided to add one as a test system, so I might have inadvertently b0rked my 12.1 VM. When I recreated the VM with 12.2 from scratch, my FreeBSD knowledge had been refreshed quite a bit so I likely set the new VM the right way.

aiosmtpd/smtp.py
```diff
@@ -162,6 +176,14 @@ class TLSSetupException(Exception):
     pass
 
 
+@public
+def sanitize(text: bytes) -> bytes:
+    m = CLIENT_AUTH_B.match(text)
+    if m:
+        return m.group("authm") + b"********" + m.group("crlf")
```
pepoluan, line 183: We can extend the `b"*"` string to 20 characters if we need more security 😝

pepoluan:
Since this is kinda urgent, after 2x24 I'll squamerge if no review.

aiosmtpd/smtp.py
```diff
@@ -436,10 +458,10 @@ async def _handle_client(self):
                     # send error response and read the next command line.
                     await self.push('500 Command line too long')
                     continue
-                log.debug('_handle_client readline: %r', line)
+                log.debug('_handle_client readline: %r', sanitize(line))
                 # XXX this rstrip may not completely preserve old behavior.
                 line = line.rstrip(b'\r\n')
-                log.info('%r >> %r', self.session.peer, line)
+                log.info('%r >> %r', self.session.peer, sanitize(line))
```
waynew, line 464: **suggestion**: Would it make more sense to create a `sanitized_log_debug`/`info`, rather than depend on remembering to call `sanitize`?

waynew (approved):
I'm :+1: on this, not sure if we could improve things a little by creating some dedicated logging functions for logging sensitive info? :shrug:

aiosmtpd/smtp.py
pepoluan: Well, that would only change the need of remembering to call `sanitize` with the need to call `sanitized_log_*`, so I don't really see the benefits.

waynew: It's definitely a small thing, but it boils down to this: 

```
log.debug('blahahahahahblaksj %s', whatever)
log.debug('bargbargbargbarg arg %s', sanitize (bob))
```

Is easy to miss that one has the sanitize call and the other is missing. Especially in a huge block of code.

It's even worse with a multitude of parameters.

Is it the end of the world? No. But are there tangible benefits that make writing and reviewing easier? Sure 😁

Basically any time you see a logging call that sends sensitive data, it should be calling the sensitive log function.

So... A small thing that I don't think should block the PR. But it would be a useful thing 🙂

pepoluan: Hmmm, you do have a point there.

Let me mull it over... I definitely don't want to make a too-complex sanitizing system... but I think I have some inkling on how to implement it nicely...

waynew: I think I was actually reviewing this first on mobile :upside_down_face: 

I'm OK if we want to merge this first and then do that in another PR. Otherwise, the thought I had during discussion is that we could either:

a) create a logger/handler implementation, so it's literally created like:

```
sanitized_logger = CustomLogger()
...
sanitized_logger.debug(...)
```

b) create a function/group of functions that live somewhere, so

```
from utils import sanitized_debug
...
sanitized_debug(...)
```

or similar.

pepoluan: I much prefer option (b), it's less complex and can be implemented cleanly.

Rather than wrapping all calls, I think of going like this:

```python
def sanitized_log(func, msg, *args):
    msg = sanitize(msg)
    gerard(log, func)(msg, *args)
```

Where sanitizing is needed, we change the call

```
log.debug(...)
```

to

```
sanitized_log("debug", ...)
```

What do you think?

waynew: Maybe just `sanitized_log(log.debug, ...)` then skip the getattr call? Otherwise I'm +1 for that approach

pepoluan: That's also good!

I'll implement that asap.

aiosmtpd/tests/test_smtp.py
```diff
@@ -36,13 +38,32 @@
...
+            encoding="latin1",
+        )
+        fhandler.setLevel(1)
+        fhandler.setFormatter(
+            logging.Formatter(
+                u"{asctime} - {name} - {levelname} - {message}",
+                style="{",
+            )
+        )
+        mail_logger.addHandler(fhandler)
```
pepoluan, line 65: If env var `AIOSMTPD_TESTLOGFILE` is set, then `SMTP` logging for this test module will also be written to that file. It's optional, but sometimes... entertainig :-)

aiosmtpd/tests/test_smtp.py
```diff
@@ -80,7 +101,14 @@ async def auth_DONT(
...
 
-    async def auth_WITH_UNDERSCORE(self, server, args):
+    async def auth_WITH_UNDERSCORE(self, server: Server, args):
+        """
+        Be careful when using this AUTH mechanism; log_client_response is set to
+        True, and this will raise some severe warnings.
+        """
+        await server.challenge_auth(
+            "challenge", encode_to_b64=False, log_client_response=True
+        )
```
pepoluan, line 111: The only reason is to test / cover the `log_client_response=True` path of `SMTP.challenge_auth()`

I was a bit lazy to write yet-another-auth-mechanism (because I'll then have to edit several places), so I just 'repurpose' existing one.

aiosmtpd/tests/test_smtp.py
```diff
@@ -241,6 +269,7 @@ class TestProtocol(unittest.TestCase):
     def setUp(self):
         self.transport = Mock()
         self.transport.write = self._write
+        self.transport.get_extra_info.return_value = "MockedPeer"
```
pepoluan, line 272: This is so that in the TESTLOGFILE, the log lines will be tidy. `MockedPeer` instead of `<Mock object bla bla bla id="12345567" etc yadda yadda>`

aiosmtpd/tests/test_smtp.py
```diff
@@ -941,6 +970,26 @@ def test_auth_good_credentials(self):
...
+                )
+        mail_logger.setLevel(oldlevel)
+        interestings = [
+            msg for msg in cm.output if "AUTH PLAIN" in msg
+        ]
+        assert len(interestings) == 2
+        assert interestings[0].startswith("DEBUG:")
+        assert interestings[0].endswith("b'AUTH PLAIN ********\\r\\n'")
+        assert interestings[1].startswith("INFO:")
+        assert interestings[1].endswith("b'AUTH PLAIN ********'")
```
pepoluan, line 991: Ensure that log is sanitized.

So in the future, if a maintainer somehow forgets that those particular lines in `smtp.SMTP` needs sanitizing, the test case will fail.

aiosmtpd/smtp.py
```diff
@@ -750,9 +785,13 @@ async def challenge_auth(
...
         #   - https://tools.ietf.org/html/rfc4954#page-13 "continue-req"
         challenge = b"334 " + (b64encode(challenge) if encode_to_b64 else challenge)
-        log.debug("Send challenge to %r: %r", self.session.peer, challenge)
+        log.debug("%r << challenge: %r", self.session.peer, challenge)
         await self.push(challenge)
         line = await self._reader.readline()
+        if log_client_response:
+            warn("AUTH interaction logging is enabled!")
+            warn("Sensitive information might be leaked!")
+            log.debug("%r >> %r", self.session.peer, line)
```
pepoluan, line 794: No sanitizing here but we do raise a pair of warnings. For _every single time_ `challenge_auth` is invoked with `log_client_response=True`.

pepoluan:
I think we're all set. I'll wait an additional 24 hours before squamerging.

aiosmtpd/tests/test_smtp.py
waynew: **praise**: I *love* providing tests for future maintainers. Usually because I'm that future maintainer :trollface:

waynew (approved):
:rocket:

pepoluan:
@waynew @warsaw I have published a new "GitHub Release". Need your help to push this to PyPI, because I'm not yet a maintainer for aiosmtpd there.

PR #248, opened 2021-02-17
Fix Controller Bug on hostname="" or IPv6 not available

### LINKED ISSUE #244: RuntimeError after upgrading from 1.2.2
After upgrading from version 1.2.2 to 1.2.4 we receive the following error when trying to start the controller:

```
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/aiosmtpd/controller.py", line 165, in start
    raise RuntimeError("Unknown Error, failed to init SMTP server")
RuntimeError: Unknown Error, failed to init SMTP server
```

This error can even be triggered by running the sample code from the [documentation](https://aiosmtpd.readthedocs.io/en/latest/controller.html):

```
import asyncio
class ExampleHandler:
  async def handle_RCPT(self, server, session, envelope, address, rcpt_options):
    return '250 OK'

controller = Controller(ExampleHandler())
controller.start()
```
After executing "controller.start()" the RuntimeError from above appears.

Python version is 3.6.8
OS is Red Hat 7.9
aiosmtpd was installed with pip3. After downgrading via "sudo pip3 install aiosmtpd=1.2.2" everything works again and aiosmtpd receives and processes emails without problems.

pepoluan:
## What do these changes do?

Previously, `Controller._testconn()`, on Linux, will raise an error if `self.hostname=""`. This is due to `socket.create_connection()` not accepting `""` as hostname parameter. (Totally no problem in Windows)

In addition, `Controller.run()` wraps _all_ exceptions generated by `_testconn()` so this particular error is not seen, leading to hard times during troubleshooting.

Also, even if `hostname` not specified, `Controller` defaults to `"::1"`, which is not available on systems with IPv6 disabled.

Finally, previously `ready_timeout` is used only to wait, with no action taken if it's reached.

So, the changes:
* Now uses a smarter way to determine 'default localhost' address by attempting `::1`, and if that fails, `127.0.0.1`
* `ready_timeout` actually acts if timeout is breached
* Only swallow `socket.timeout` error but raises others

User need to be aware that `ready_timeout` is no longer lip service; it is actually enforced with a possibility of raising a `TimeoutError` if violated.

Other than that, there should be no breaking behavioral changes.

## Related issue number

Fixes #244 

## Checklist

   - [x] Windows 10 (via Cygwin): `qa,py-{nocov,cov}`
   - [x] Ubuntu 18.04 on WSL 1.0: `(ALL+pypy3)`
   - [x] FreeBSD 12.2 on VBox: `(ALL+pypy3)`

### DIFF
--- a/aiosmtpd/controller.py
+++ b/aiosmtpd/controller.py
@@ -5,10 +5,11 @@ import asyncio
 import errno
 import os
 import ssl
-import time
 import threading
+import time
 from contextlib import ExitStack
-from socket import create_connection
+from socket import AF_INET6, SOCK_STREAM, create_connection, has_ipv6
+from socket import socket as makesock
 from socket import timeout as socket_timeout
 from typing import Any, Coroutine, Dict, Optional
 from warnings import warn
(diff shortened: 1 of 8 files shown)


pepoluan:
Blergh MacOS tests are so flakey ...

Should I just drop them?

pepoluan:
> 
> 
> Blergh MacOS tests are so flakey ...
> 
> Should I just drop them?

See? I changed nothing, I just "re-run workflow" and everything now passes.

Damn MacOS 😑

pepoluan:
ALrightey ... because this is a kinda-urgent bugfix, I won't let this marinate for too long.

Come tomorrow I'll squamerge this fix and push a release.

release.py
```diff
@@ -45,8 +46,21 @@
...
+with newsfile.open("rt") as fin:
+    want = re.compile("^" + re.escape(version) + r"\s*\(\d{4}-\d\d-\d\d\)")
+    for ln in fin:
+        m = want.match(ln)
+        if not m:
+            continue
+        break
+    else:
+        print(f"ERROR: I found no datestamped entry for {version} in NEWS.rst!")
+        sys.exit(1)
```
pepoluan, line 60: Thankfully I've never actually _released_ a version with datestamp still `aiosmtpd-next`. I think.

But just in case, let's enforce this.

pepoluan (commented):
Some comments on possibly interesting decisions I made.

aiosmtpd/controller.py
```diff
@@ -17,6 +20,24 @@
...
+        # And since it worked, it's clear that IPv6 is supported
+        return "::1"
+    except OSError as e:
+        if e.errno == errno.EADDRNOTAVAIL or e.errno == 10049:
+            # 10049 is WSAEADDRNOTAVAIL .. seems to only exist on Windows
+            # So we'll not use an errno constant there.
+            return "127.0.0.1"
+        if e.errno == errno.ECONNREFUSED:
+            return "::1"
+        raise
```
pepoluan, line 38: Detect "::1" by trying to open a connection to it. If successful, return "::1". If not, let's analyze the exception and respond accordingly.

PR #267, opened 2021-04-01
Explicitly require setuptools, tests/test_smtps.py and tests/test_starttls.py imports pkg_resources

hrnciar:
## What do these changes do?

Some files have runtime dependency on setuptools because of importing `pkg_resources` module. In Fedora, we build the package using setuptools, but we ship it to users potentially without it. I suggest explicitly list this requirement. 

No

## Checklist

  - [x] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py36,py37,py38,py39}-{nocov,cov,diffcov}, qa, docs`
  
### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,12 +35,15 @@ zip_safe = false
 python_requires = ~=3.6
 packages = find:
 include_package_data = true
+setup_requires =
+    setuptools
 install_requires =
     atpublic
     attrs
     typing-extensions ; python_version < '3.8'
 tests_require =
     tox
+    setuptools
 
 [options.packages.find]
 exclude =


setup.cfg
```diff
@@ -38,6 +38,7 @@ include_package_data = true
 install_requires =
     atpublic
     attrs
+    setuptools
```
Conan-Kudo, line 41: This should probably be a `tests_require` instead of `install_requires`.

Conan-Kudo (changes requested):

setup.cfg
waynew: I'm not sure that's correct - we use setuptools on install... Why wouldn't that be a dep?

waynew (approved):

setup.cfg
Conan-Kudo: `install_requires` is for it to be on the system at runtime. `setup_requires` is for installation itself, and you have a `tests_require` for dependencies for tests.

waynew: Then it would be more appropriate it a setup_requires, no?

Conan-Kudo: Yep, you're right!

pepoluan:
Closing and reopening to trigger GHCI

pepoluan:
All green! LGTM!

PR #270, opened 2021-05-12
Implement custom SSL handshake timeout on Controller

### LINKED ISSUE #269: Allow setting ssl_handshake_timeout on a Controller
The 60 seconds default timeout in `sslproto.SSLProto` for the SSL handshake to complete is longer than necessary for many scenarios (e.g. testing against a local SMTP server) with aiosmtpd.

If a client does not complete the SSL handshake the current behaviour is that the connection will close after the `ssl_handshake_timeout` of 60 seconds on python 3.7 and later.  This raises a `smtplib.SMTPServerDisconnected` exception. The following test (which assumes that it is included in test_smtps.py) is an example of triggering this using the `smtplib.SMTP` client (which never completes the handshake):

```python
# in ./aiosmtpd/tests/test_smtps.py

@pytest.mark.skipif(sys.version_info < (3, 7),
                    reason="SSL timeout implemented implemented with 3.7")
def test_SSL_timeout(self, ssl_controller):
    with pytest.raises(SMTPServerDisconnected) as ex,\
         SMTP(*Global.SrvAddr) as smtp_client:
        # smtplib.SMTP does not support opporutnistic SSL so the SSL
        # handshake never completes. On Python 3.6 and earlier this means
        # that the connection will hang.
        smtp_client.helo("example.com")
    assert "Connection unexpectedly closed" in str(ex)
```

To partially mitigate the effect of this the `ssl_handshake_timeout` parameter on the `loop.create_server()` can be set to a shorter time in environments that don't need the full minute to complete. Unfortunately this is available only as of python 3.7; I have been unable to find a reasonable solution that would work for python 3.6.

My proposal is:
+ Add a parameter to the `Controller` classes for the `ssl_handshake_timeout` that is passed to `loop.create_server()` if using python 3.7 or later.
+ Use the same timeout for STARTTLS on python 3.7 or later.
+ Warn on python 3.6 (or earlier for completeness even if not supported) that this has no effect.
+ Do not add an argument for the SSL handshake timeout to the CLI until python 3.6 support is dropped from aiosmtpd at some point in the future.

bebleo:
## What do these changes do?

This implements a custom `ssl_handshake_timeout` parameter on Controllers that is then passed to the `loop.create_server()` and used in the `smtp_STARTTLS` methods to modify the amount of time that the server will wait for the SSL handshake to complete.

## Related issue number

Closes #269 

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: "42 17 * * 1"
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ python ]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
+          queries: +security-and-quality
+
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v2
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
+        with:
+          category: "/language:${{ matrix.language }}"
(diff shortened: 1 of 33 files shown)


pepoluan:
If the conflicts can be resolved, this can go into 1.4.5

bebleo:
Cleaning this up and rebasing it to the latest release will take more time than starting anew. I'll close this for now, but leave the issue open for consideration in a future release.

Issue #272, opened 2021-06-06
Controller throws a TimeoutError when binding a unspecified address


wevsty:
Controller throws a TimeoutError when binding a unspecified address.
Affected versions: aiosmtpd (1.2.3-1.4.2)
Example:
```python
from aiosmtpd.controller import Controller

class CustomHandler:
    async def handle_DATA(self, server, session, envelope):
        return '250 OK'

if __name__ == '__main__':
    handler = CustomHandler()
    controller = Controller(handler, hostname='0.0.0.0', port=10025)
    controller.start()
    input('SMTP server running. Press Return to stop server and exit.')
    controller.stop()
```
Running this code will prompt an error
```python
Traceback (most recent call last):
  File "C:\USER_PROGRAM\PyCharm\plugins\python-ce\helpers\pydev\pydevd.py", line 1483, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "C:\USER_PROGRAM\PyCharm\plugins\python-ce\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/root/PycharmProjects/Projects/test.py", line 10, in <module>
    controller.start()
  File "C:\Users\root\PycharmProjects\Projects\venv\lib\site-packages\aiosmtpd\controller.py", line 231, in start
    raise TimeoutError(
TimeoutError: SMTP server started, but not responding within allotted time. This might happen if the system is too busy. Try increasing the `ready_timeout` parameter.
```
This is obviously a clear mistake. That needs to be fixed as soon as possible.

1756816846:
Why is there such a mistake

Traceback (most recent call last):
  File "C:/Users/Darkness/Desktop/crazy-email-recv-srv-master/dd.py", line 11, in <module>
    controller.start()
  File "D:\python37\lib\site-packages\aiosmtpd\controller.py", line 315, in start
    "SMTP server started, but not responding within allotted time. "
TimeoutError: SMTP server started, but not responding within allotted time. This might happen if the system is too busy. Try increasing the `ready_timeout` parameter.

PR #273, opened 2021-06-06
Fix bind unspecified address controller will raise TimeoutError

wevsty:
## What do these changes do?

Add a function is_unspecified_address() determine if the address is the unspecified address.
Automatic replacement of unspecified address with local addresses.

This will be more compatible with the user-specified behavior.

## Related issue number

issue #272

## Checklist


waynew:
I'm not 100% sure if this is the correct fix -  0.0.0.0 should bind on all interfaces right? I'm on my phone atm so I might not grok correctrly...

wevsty:
> 
> 
> I'm not 100% sure if this is the correct fix - 0.0.0.0 should bind on all interfaces right? I'm on my phone atm so I might not grok correctrly...

Listening ```0.0.0.0``` usually means listening to all IPV4 interfaces.
However, ```0.0.0.0``` is not directly accessible, so there are errors that need special handling when detecting it.
On IPV6 ```::``` has the same problem

wevsty:
I did my best to fix the problems with flake8 checks.
On my machine I passed all the tests.

But ```test_byclient``` still doesn't pass the test in some environments,This should be due to a change in Python's behavior.
see #219 
I think this place needs to make some changes @pepoluan

waynew:
@pepoluan Any reason we shouldn't merge this?

pepoluan:
> @pepoluan
> 
> > Shall we LGTM?
> 
> All checks does not seem to find any problems. If there are other questions please contact me.

After one year (sorry) conflicts arise ...

Can we fix this and revisit?

wevsty:
> @pepoluan

It's been a bit of a busy couple of weeks, so I've procrastinated a bit.
But I've resolved the conflicts submitted the new version and it should work fine now.

pepoluan:
Okay so I've drunk enough coffee for today, I can think again 😁

This converts "unspecified" to localhost, rather than current IP adress, I think we need to mention that in the documentation as well. Other software when given "unspecified" will bind to all existing addresses including localhost. Though that is not a 'standard behavior' I think being explicit will be good.

wevsty:
> Okay so I've drunk enough coffee for today, I can think again 😁
> 
> This converts "unspecified" to localhost, rather than current IP adress, I think we need to mention that in the documentation as well. Other software when given "unspecified" will bind to all existing addresses including localhost. Though that is not a 'standard behavior' I think being explicit will be good.

I don't think it needs to be deliberately stated in the documentation.
Bind unspecified address is a common behavior, and on many platforms unspecified address is used instead of the IP address of all interfaces on the local machine.
And the code is not replacing bind 0.0.0.0 with 127.0.0.1.
It's just that aiosmtpd tries to connect to the bind address after the bind is done to make sure the bind is successful, but obviously most systems don't accept 0.0.0.0 as a connectable address, this PR just uses 127.0.0.1 when bind 0.0.0.0 to for connection testing, and does not change the external behavior.

waynew:
> I don't think it needs to be deliberately stated in the documentation.
> Bind unspecified address is a common behavior, and on many platforms unspecified address is used instead of the IP address of all interfaces on the local machine.
> And the code is not replacing bind 0.0.0.0 with 127.0.0.1.


It may be common behavior, but I only learned about this just recently. [I'm sure I'm not the only one](https://xkcd.com/1053/)

pepoluan:
> And the code is not replacing bind 0.0.0.0 with 127.0.0.1.
> It's just that aiosmtpd tries to connect to the bind address after the bind is done to make sure the bind is successful, but obviously most systems don't accept 0.0.0.0 as a connectable address, this PR just uses 127.0.0.1 when bind 0.0.0.0 to for connection testing, and does not change the external behavior.

Ahh thanks for the explanation!

Was a bit confused when I just look at the snippets.

PR #280, opened 2021-08-04
fixed right order for EHLO handler

hudecof:
## What do these changes do?

Fix in EHLO handler, where the `250 HELP` response line, means without dash should be the last one.
My tetsing postix server do not accept nre EHLO extensions after `250 HELP`

There is no change for user, just fix according the RFC https://datatracker.ietf.org/doc/html/rfc5321#section-4.1.1.1

## Related issue number

None

## Checklist

- [x ] I think the code is well written

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -854,8 +854,8 @@ class SMTP(asyncio.StreamReaderProtocol):
             # New behavior: hand over list of responses so far to the hook, and
             # REPLACE existing list of responses with what the hook returns.
             # We will handle the push()ing
-            response.append('250 HELP')
             response = await self._call_handler_hook("EHLO", hostname, response)
+            response.append('250 HELP')
 
         for r in response:
             await self.push(r)


waynew:
I need to double check this - I *think* it's the right fix, I think I may have overridden this method within my own code for that same reason.

waynew:
Confirmed that yes, this is the right fix.

However, we don't have a test failing for this case, so we really should create that test case. @hudecof is that something that you could tackle this week or next?

hudecof:
> HELP

hudecof:
I did not want to close this ;)
I could try, to write the test, but bit sure about the time line ;(

pepoluan:
Running the GHCI to test correctness.

Until we create the test to ensure this won't happen again, I won't merge it though.

I hope I can find some time to write the necessary tests if you're occupied @hudecof

pepoluan:
This still use the old GHCI definition file, resulting in very slow qa_docs (and usually ends up in failure)

If you can rebase this to the latest master, @hudecof , I can rerun the tests.

Though again, until tests are written to prevent same issue from happening, I won't merge this.

But at least we can be sure that the changes do not cause any regression.

PR #292, opened 2021-11-21
Fix SMTPS / implicit TLS

### LINKED ISSUE #274: Implicit TLS (SMTPS) support is botched
When using --smtpskey and --smtpscert, after EHLO, AUTH fails with

```
538 5.7.11 Encryption required for requested authentication mechanism
```

Either I am, or this is clearly wrong.
I suggest that it awaits starttls, just as when people use explicit TLS.

Am I missing something?
If I am to send a PR, how does one run tests here?

### LINKED ISSUE #281: AUTH not provided in EHLO with SSL/TLS
When using implicit SSL/TLS (so no STARTTLS), the `self._tls_protocol` will never be set. This results in `AUTH` not being present in the EHLO response (given that `auth_require_tls` is true): https://github.com/aio-libs/aiosmtpd/blob/master/aiosmtpd/smtp.py#L833

I would expect that when an SSL/TLS connection is used instead of STARTTLS (and `auth_require_tls` is true), `AUTH` is provided to the client. Currently, we are setting `auth_require_tls` is false to circumvent this issue.

strongholdmedia:
## What do these changes do?

Allow for implicit TLS usage (based solely on provided context data as, like @pepoluan mentioned in #281, it is not quite feasible to detect any wrapping (nor should it be done).

Should not be any, as the only change affects a feature that did not quite work, like, ever.

## Related issue number

Fixes #274 and most likely also closes #281.

## Checklist

- [✓] I think the code is well written
- [  ] Unit tests for the changes exist
- [  ] tox testenvs have been executed in the following environments:
  
  - [  ] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py36,py37,py38,py39}-{nocov,cov,diffcov}, qa, docs`
  - [  ] Windows (7, 10): `{py36,py37,py38,py39}-{nocov,cov,diffcov}`
  - [  ] WSL 1.0 (Ubuntu 18.04): `{py36,py37,py38,py39}-{nocov,cov,diffcov}, pypy3-{nocov,cov}, qa, docs`
  - [  ] FreeBSD (12.2, 12.1, 11.4): `{py36,pypy3}-{nocov,cov,diffcov}, qa`
  - [  ] Cygwin: `py36-{nocov,cov,diffcov}, qa, docs`
- [✓] Documentation reflects the changes
- [✓] Add a news fragment into the `NEWS.rst` file
  
  * I cannot find "NEWS.rst" in this repository

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -322,6 +322,7 @@ class SMTP(asyncio.StreamReaderProtocol):
             ident: str = None,
             tls_context: Optional[ssl.SSLContext] = None,
             require_starttls: bool = False,
+            is_using_smtps: bool = False,
             timeout: float = 300,
             auth_required: bool = False,
             auth_require_tls: bool = True,
@@ -368,11 +369,13 @@ class SMTP(asyncio.StreamReaderProtocol):
         self.envelope: Optional[Envelope] = None
         self.transport = None
         self._handler_coroutine = None
-        if not auth_require_tls and auth_required:
+        if not auth_require_tls and not is_using_smtps and auth_required:
             warn("Requiring AUTH while not requiring TLS "
                  "can lead to security vulnerabilities!")
-            log.warning("auth_required == True but auth_require_tls == False")
+            log.warning("auth_required == True but auth_require_tls == False "
+                        "and not using SMTPS")
         self._auth_require_tls = auth_require_tls
+        self._is_using_smtps = is_using_smtps
 
         if proxy_protocol_timeout is not None:
             if proxy_protocol_timeout <= 0:
@@ -917,7 +920,7 @@ class SMTP(asyncio.StreamReaderProtocol):
             return
         elif not self.session.extended_smtp:
             return await self.push("500 Error: command 'AUTH' not recognized")
-        elif self._auth_require_tls and not self._tls_protocol:
+        elif self._auth_require_tls and not self._tls_protocol and not self._is_using_smtps:
             return await self.push("538 5.7.11 Encryption required for requested "
                                    "authentication mechanism")
         elif self.session.authenticated:
--- a/aiosmtpd/main.py
+++ b/aiosmtpd/main.py
@@ -241,6 +241,13 @@ def main(args: Optional[Sequence[str]] = None) -> None:
     else:
         tls_context = None
 
+    if args.smtpscert and args.smtpskey:
+        smtps_context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+        smtps_context.check_hostname = False
+        smtps_context.load_cert_chain(str(args.smtpscert), str(args.smtpskey))
+    else:
+        smtps_context = None
+
     factory = partial(
         SMTP,
         args.handler,
@@ -248,6 +255,7 @@ def main(args: Optional[Sequence[str]] = None) -> None:
         enable_SMTPUTF8=args.smtputf8,
         tls_context=tls_context,
         require_starttls=args.requiretls,
+        is_using_smtps=smtps_context is not None,
     )
 
     logging.basicConfig(level=logging.ERROR)
@@ -260,13 +268,6 @@ def main(args: Optional[Sequence[str]] = None) -> None:
         log.setLevel(logging.DEBUG)
     if args.debug > 2:
         loop.set_debug(enabled=True)
-
-    if args.smtpscert and args.smtpskey:
-        smtps_context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
-        smtps_context.check_hostname = False
-        smtps_context.load_cert_chain(str(args.smtpscert), str(args.smtpskey))
-    else:
-        smtps_context = None
 
     log.debug("Attempting to start server on %s:%s", args.host, args.port)
     server = server_loop = None


waynew:
I would like to see a test that can confirm this behavior - ideally one that we could run in the automated tests, but I'd even be OK with one that we could run out of band :thinking: 

@strongholdmedia are you able to do that in the next week or two?

strongholdmedia:
@waynew well, my schedule is unusually overcrowded for the end of the year, but I can probably manage it in say, three weeks, if not two.
But I would like some encouragement about it getting processed in a shorter time frame – given that I created this PR just a bit over a year ago :)

waynew:
No worries - we've definitely seem some ebb in our activity 😅 obviously I'm getting back into the flow here.

If you *can* get it, that would be awesome - but I realize that life might be happening. Either way, I'm going to do what I can to get all these changes in before I cut a release.

pepoluan:
Ohhh this is nice ... but a bit complex-y...

I really want to merge this, but let's do it after 1.4.3 goes final.

We're trying to ensure projects that depends on aiosmtpd to have time to test with 1.4.3 and make the Debian Freeze (see #322 )

strongholdmedia:
> We're trying to ensure projects that depends on aiosmtpd to have time to test with 1.4.3 and make the Debian Freeze (see #322 )

I still don't really understand this issue.
For once, this should be 100% backwards compatible (and, per the previous tests, it is).
On the other hand, it is 2022; we (and presumably wherever the providers don't have mandatory government intervention/rigging _nobody_ else) are not using _anything but implicit TLS_ for over 5-6 years.

strongholdmedia:
@waynew I checked stuff and it seems that a very basic test case is provided historically for SMTPS.

I tried to extend it somewhat, for the mere sake of activity :) but I have a hard time figuring out tox's "internal error" that it throws on me during coverage testing (with the main branch as well, so nothing specific to the changes).

On the other hand, the basic test case may already suffice, as:

* as per the RFC, there is _no way_ to distinguish between implicit TLS and plaintext protocol within the wrapped handler (here, smtpd itself)
* the SSL/TLS context already having tests in the STARTTLS test
* as such, if the SMTP test runs properly, and the STARTTLS as well (so that the SSL context handling is okay), the only thing that is left is to check whether the client can (still) communicate with the server when smtps options are provided; and that is covered by the _skeletal_ test case being included since who knows when

As such, I am not quite sure what to test further – as you either have TLS wrapping, and everything _appears as if it was_ plaintext, or you have no TLS wrapping when you expect it, and you get random "binary" garbage.

Meanwhile I merged upstream changes so that it is not outdated any further.

strongholdmedia:
Ah, now I see that there should probably be a test case for #281 (that is, having AUTH in the EHLO response).

strongholdmedia:
I checked smtplib docs and the test should work as it is. But I won't know until the workflow is authorized, as sadly, I must be missing something with the tox environments.
Will check tox sources for what could be causing the issue for me tomorrow.
Meanwhile, I hope the test will run :)

strongholdmedia:
I did, thanks. The previous test used the hardcoded response string for comparison, but the EHLO method returns a tuple IIRC.

pepoluan:
> I did, thanks. The previous test used the hardcoded response string for comparison, but the EHLO method returns a tuple IIRC.

The previous test actually compares the tuple returned to a well-known tuple 'constant' in another file. Just because I get tired of checking both the code and the string returned separately, every time 😄

pepoluan:
Coming nearer to fixing this! Only 2 failed tests actually, though spread across all the OSes 😆

Fail #1
```
FAILED aiosmtpd/tests/test_smtp.py::TestAuthArgs::test_warn_authreqnotls - AssertionError: assert ('mail.log', ... using SMTPS') == ('mail.log', ...tls == False')
  At index 2 diff: 'auth_required == True but auth_require_tls == False and not using SMTPS' != 'auth_required == True but auth_require_tls == False'
```
Will need to change the expected string 😊

Fail #2
```
FAILED aiosmtpd/tests/test_smtps.py::TestSMTPS::test_smtps - AssertionError: assert False
 +  where False = <bound method SMTP.has_extn of <smtplib.SMTP_SSL object at 0x7fca56664c10>>('AUTH')
```
This might need a bit deeper digging; seems like `has_extn('AUTH')` is not returning `True`?

strongholdmedia:
> **Edit:** I think you need to drop line `test_smtps.py#L45`: The SMTP class being instantiated does NOT have an authenticator. After all, the test simply tests that SMTP works inside an implicit SSL tunnel.

Well, the complaint in #281 was (duly so) that AUTH is not provided when using SMTPS. (I.e. handled as plaintext.)
That was definitely not optimal from any perspective.

This is now definitely fixed, I tested it manually yesterday (running the module directly).

Currently _if_ a "ssl" prop (a TLS context) is provided to `create_server`, it will use SMTPS. (I think this was sort of historical behavior, it just never worked.)

So that I need to figure out how to pass a pre-wired TLS context to the constructor within the test (i.e. where to get key and cert from), then it should work.

waynew:
aiosmtpd/tests/certs/server.crt should be the cert you're looking for 👍

pepoluan:
> aiosmtpd/tests/certs/server.crt should be the cert you're looking for 👍

Aren't we supposed to refresh that cert every now and then?

strongholdmedia:
> Aren't we supposed to refresh that cert every now and then?

I see it is valid until `May  8 13:55:26 2119 GMT`.
So in theory yes, but I guess it will works for some more weeks coming :)

strongholdmedia:
I missed the fact that all "extra" arguments are passed to the factory, then the controller, and finally to `SMTP` class.
This should be the proper approach – unsure if it really works though.

pepoluan:
Wow, this time the change caused all jobs to fail. All jobs fail at exactly the same test case:

```
FAILED aiosmtpd/tests/test_smtps.py::TestSMTPS::test_smtps - AssertionError: assert 'smtps' in {'8bitmime': '', 'auth': ' LOGIN PLAIN', 'help': '', 'size': '33554432', ...}
 +  where {'8bitmime': '', 'auth': ' LOGIN PLAIN', 'help': '', 'size': '33554432', ...} = <smtplib.SMTP_SSL object at 0x7f5325683510>.esmtp_features
```

strongholdmedia:
Well, my bad, it was not a change now but 2 days ago, which was some sort of leftover attempt. :/ But removed it now.
To up the challenge, added a line that tests that _STARTTLS is not available_, since that would make no sense to wrap the already encrypted communication.

waynew:
> > aiosmtpd/tests/certs/server.crt should be the cert you're looking for 👍
> 
> Aren't we supposed to refresh that cert every now and then?

It expired a year or two ago, breaking all the tests. So I went ahead and generated one that was good for 100 years, since it's only used in testing.

Next time it needs help it won't be my problem :trollface:

strongholdmedia:
@pepoluan @waynew any news here? :) Not to be impatient or something, but I suggest the "in progress" tag be removed, and also the "enhancement" is debatable, as this is actually for a core feature that never worked according to the RFC (even the test run blank for it).

pepoluan:
> @pepoluan @waynew any news here? :) Not to be impatient or something, but I suggest the "in progress" tag be removed, and also the "enhancement" is debatable, as this is actually for a core feature that never worked according to the RFC (even the test run blank for it).

At the moment I'm really concerned with the deprecation of `get_event_loop()` which had passed under my radar. If not satisfactorily handled, this will cause everything relying on aiosmtpd to grind to a halt.

So I'm prioritizing #355 first.

aiosmtpd/smtp.py
```diff
@@ -322,6 +322,7 @@ def __init__(
             ident: Optional[str] = None,
             tls_context: Optional[ssl.SSLContext] = None,
             require_starttls: bool = False,
+            is_using_smtps: bool = False,
```
pepoluan, line 325: After doing some tracing, a thought occurred to me:

Implicit TLS (SMTPS) is triggered if `tls_context` is set, right?

Can we just set `is_using_smtps` attribute from that? So `self._is_using_smtps = (tls_context is not None)`?

I might've overlooked something, so I welcome a discussion here.

pepoluan (commented):
Some thoughts, discussions welcome.

pepoluan:
Moved this to 1.4.4 milestone because it doesn't seem to cause any b0rkages.

However, refraining from merging until the 'discussion' above has been resolved.

aiosmtpd/smtp.py
strongholdmedia: Well, yes and no.
For once, `tls_context` is for `STARTTLS`; `SMTPS` uses `ssl_context`. (This was already such before, I did not change it.)

Otherwise we _could_ set that, yep, but this option communicates _clear intent_.

As you've noted in #281, there is _no possible means for detecting_ whether or not we are using implicit TLS (`SMTPS`) within the `SMTP` class.

This goes to the extent that `ssl_context` is actually _not used by the `SMTP` class **at all**_.
It is still kept by me, as if later authentication methods will require e.g. client re-validation or something, it could happen.

But unlike `is_using_smtps`, `ssl_context` may or may not be an explicit intent.
`ssl_context` being `None` could equally mean "should not use SMTPS", as well as "hey, we/they/... failed to handle a key / cert error / expiry / ... somewhere".

But as per the above, there is _no means_ to detect whichever from inside the `SMTP` class; as such, if the described scenario happens, the client/user would just note that "hey, we don't have `AUTH`", or even "got `SMTP` error, pls help, it worked yesterday".

Currently, if `is_using_smtps` is set with no `ssl_context`, it should present an error.
I understand your point and will change it if that'd be the agreement, but I feel that would be a technically less correct approach.

ThiefMaster:
Could this please be merged and a fixed version released? :)

PR #311, opened 2022-10-31
tests: Run pytype as a part of tox qa testenv

maxking:
## What do these changes do?

Run type checking during the tox qa env so that we can test
locally for any typing related errors instead of manually
installing and running pytype.

## Are there changes in behavior for the user?

Developers will now also see typing related failures when
running qa tox env locally.

## Related issue number

Fixes #310 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] tox testenvs have been executed in the following environments:
  
  - [ ] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py36,py37,py38,py39}-{nocov,cov,diffcov}, qa, docs`
  - [ ] Windows (7, 10): `{py36,py37,py38,py39}-{nocov,cov,diffcov}`
  - [ ] WSL 1.0 (Ubuntu 18.04): `{py36,py37,py38,py39}-{nocov,cov,diffcov}, pypy3-{nocov,cov}, qa, docs`
  - [ ] FreeBSD (12.2, 12.1, 11.4): `{py36,pypy3}-{nocov,cov,diffcov}, qa`
  - [ ] Cygwin: `py36-{nocov,cov,diffcov}, qa, docs`
- [ ] Documentation reflects the changes
- [ ] Add a news fragment into the `NEWS.rst` file
  
  * Add under the "aiosmtpd-next" section, creating one if necessary
    * You may create subsections to group the changes, if you like
  * Use full sentences with correct case and punctuation
  * Refer to relevant Issue if applicable

PR #312, opened 2022-10-31
Add vim swap files to gitignore

waynew:
What it says on the tin.

Was working on reviewing #311 and the qa tools complained, loudly, about my swap files :joy:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ diffcov-*.html
 prof/
 .pytype/
 ~temp*
+*.sw[a-p]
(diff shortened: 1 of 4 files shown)


waynew:
well, looks like that wouldn't actually help, doh! - though I'm wondering if I can ignore them in the sdist:

> lists of files in version control and sdist do not match!

waynew:
Looks like there's a bit more - I need to add the ignores to MANIFEST.in as well. I was also still seeing the QA fail - was failing on the news/version last check. I'll need to poke at that a bit more, too :thinking:

waynew:
Okay, *this* works now.

waynew:
Okay, assuming all the tests pass, this should be good to merge whenever. Will probably merge tomorrow afternoon if there are no objections.

waynew:
No objections, so merging

PR #323, opened 2022-12-14
Faster GitHub CI

pepoluan:
## What do these changes do?

Speeds up the qa portion of GitHub CI.

## Are there changes in behavior for the user?

None.

## Related issue number

None.

Issue #322, opened 2022-12-11
Python 3.11 compatible release


CharlemagneLasse:
It seems like various distributions will loose mailman 3 because they switched to python 3.10 or 3.11 - and the last release of aiosmtpd doesn't have support for these python versions. See #309 and #314

For example, Debian will have its first freeze in January 2023: https://release.debian.org/bookworm/freeze_policy.html
It would be nice when there is a new release (after newly 2 years) which is compatible with the current Python versions.

CharlemagneLasse:
@pepoluan, the Debian bug about aiosmtpd removal can be found at https://bugs.debian.org/1025019

pepoluan:
Okay, that's serious.

What is the root cause of the test failures?

CharlemagneLasse:
> What is the root cause of the test failures?

Afaik, that Debian packages the latest released version. And the latest released version doesn't include the fixes from #309 + #314.

pepoluan:
Hmmm.... so basically we better package the latest as, say, 1.5.0 and ask the Debian maintainers to just go forward with 1.5.0?

CharlemagneLasse:
That would be my hope. But of course, I can only make suggestions. You know the project and its state - but I thought it is fair to make this suggestion because Debian is not the only distribution facing such problems - and users of pypi also seem to be unable to use aiosmtpd with the current Python version (due to the age of the last release)

pepoluan:
I have just fixed GitHub CI. All tests run green now. I am 90% confident of packaging into 1.4.3.

P-EB:
Thanks for the work. I'm happy to package in Debian either 1.4.3 or 1.5.0 if you manage to finish it by mid-january. :)

Anyway, I'll take the first release you shoot.

pepoluan:
> Thanks for the work. I'm happy to package in Debian either 1.4.3 or 1.5.0 if you manage to finish it by mid-january. :)
> 
> Anyway, I'll take the first release you shoot.

Well, for 1.4.3 I think this month can be released. The changes are minor anyways.

1.5.0 has many new ideas and will represent quite a heavy lifting, so it might not be able to make it even in mid-January 😅 

I'm currently running the GitHub CI tests again to ensure that things are still green, then gonna update NEWS.rest ... and then figure out how I used to release the packages 😆

pepoluan:
Version 1.4.3rc1 released.

[...]

PR #324, opened 2022-12-14
Enhance GitHUB CI Workflow

### LINKED ISSUE #316: RFE: Update the test matrix for aiosmtpd
We should probably update the supported OSes and Python versions in the CI. Here is what I propose:

* Bump Ubuntu to 22.10 (latest version) and drop other one.
* Bump macOS to 12 (13 _just_ came out, so we can possibly bump to that after a while?)
* Drop support for Python 3.6 and pypy3.6
* Add support for Pypy3.7-3.9

With addition of 3.11 (see #314) this would make the test matrix *huge* though with 5 CPythons and 3 Pypy and 3 OSes (8x3=24 runs, not including LGTM and Qa) 😄 

Depending on what folks think, we _could_ limit to using only 1 ubuntu version (as representation for Linux) and maybe not run _every_ python on _every_ OS. Although, Idk how much customization does GHA matrix declarations support to even do that.

pepoluan:
Speedup of #323 + many modernizations, such as:

* Add pypy3.8
* Exclude pypy3.6 on Windows (no longer available)
* Remove 3.6 (end of life already -- let it rest in peace)
* Actions use Node 16

And a workaround:
* Skip code coverage when running on pypy3.8 (not sure why it doesn't hit 100%)

Now all steps are green and runs in acceptable time, as can be seen in [CI #451](https://github.com/aio-libs/aiosmtpd/actions/runs/3696059951)

If merged this closes #316

### DIFF
--- a/.github/workflows/unit-testing-and-coverage.yml
+++ b/.github/workflows/unit-testing-and-coverage.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           # 3.8 is chosen because it seems to be the fastest for <3.9
           # (3.9 excluded because it seems to be still very unstable)
-          python-version: 3.11
+          python-version: 3.8
       - name: "Update modules & install deps"
         # language=bash
         run: |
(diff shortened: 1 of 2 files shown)


pepoluan:
When this is merged, we can again be sure that our CI no longer raises false alarm, and I can package & tag 1.4.3.

pepoluan:
If no objections within say 2x24 hours I will squamerge, then package the code so far into 1.4.3.

Hopefully can answer the "severe/critical" blocker Debian maintainer are facing right now (as mentioned in #322 )

pepoluan:
Testing with the latest commit 752fae9 : [All is green](https://github.com/aio-libs/aiosmtpd/actions/runs/3710376932)

Can squamerge this.

Issue #334, opened 2022-12-20
Drop older asyncio workarounds for Python < 3.7


pepoluan:
Since we've dropped 3.6, and 3.7 has A LOT of asyncio improvements ([here's a glance of some](https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-asyncio) -- already a long list!), we might want to start replacing all 'workarounds' that catered to Python < 3.7 to things that are 'more elegant' (and definitely more up-to-date).

Not marking this for any particular version. But definitely making this a tech-debt.

pepoluan:
With PR #359 , we have completely removed any support for Python<3.7, and this issue can go forward.

PR #355, opened 2022-12-28
Workaround for asyncio.get_event_loop() deprecation

### LINKED ISSUE #353: Do not call `asyncio.get_event_loop()` if no existing event loop
As described in https://docs.python.org/3/library/asyncio-eventloop.html , directly calling `asyncio.get_event_loop()` raises DeprecationWarning if no event loop is already running. The behavior of `get_event_loop()` was changed in Python 3.10.

Up to Python 3.9, `get_event_loop()` will silently create an event loop with no warnings.

_It seems that_ in Python 3.10 and 3.11, `get_event_loop()` will still create an event loop, but a DeprecationWarning is raised.

The `get_event_loop()` behavior will change completely in Python 3.12, no longer creating an event loop but will raise an Error instead.

We must parse the code and ensure that `get_event_loop()` expecting an implicit event loop creation are removed from the code.

pepoluan:
## What do these changes do?

Implement workaround due to the deprecation in ***behavior*** of asyncio.get_event_loop() (which will outright result in error with Python 3.12)

This change of behavior first appeared in Python 3.10: https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop

And there has been some discussion on this, among others:
* https://github.com/python/cpython/issues/83710
* https://github.com/python/cpython/issues/93453
* https://github.com/python/cpython/pull/98440

Should be minimal.
* If user creates their own event loop and feeds that to aiosmtpd, then nothing will be changed
* If user does not create their own event loop and let aiosmtpd generates one for them, then the behavior totally depends on how `get_event_loop_policy()` was in effect at time of instantiation.
  *  This might or might not be a problem. For asyncio, the behaviour should be similar. I am not sure about the behavior of other implementations (e.g., `uvloop`)

## Related issue number

Closes #353 

## Checklist

  - [x] Windows (10): `py311-{nocov,cov}`

### DIFF
--- a/aiosmtpd/__init__.py
+++ b/aiosmtpd/__init__.py
@@ -4,7 +4,7 @@ import asyncio
 import warnings
 
 
-__version__ = "1.4.4a0"
+__version__ = "1.4.4a1"
 
 
 def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
--- a/aiosmtpd/__init__.py
+++ b/aiosmtpd/__init__.py
@@ -1,4 +1,18 @@
 # Copyright 2014-2021 The aiosmtpd Developers
 # SPDX-License-Identifier: Apache-2.0
+import asyncio
+import warnings
+
 
 __version__ = "1.4.4a0"
+
+
+def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        try:
+            loop = asyncio.get_event_loop()
+        except (DeprecationWarning, RuntimeError):
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+    return loop
--- a/aiosmtpd/__init__.py
+++ b/aiosmtpd/__init__.py
@@ -12,7 +12,7 @@ def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
         warnings.simplefilter("error")
         try:
             loop = asyncio.get_event_loop()
-        except (DeprecationWarning, RuntimeError):  # pragma: no cover
+        except (DeprecationWarning, RuntimeError):  # pragma: py-lt-310
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
     return loop
--- a/aiosmtpd/__init__.py
+++ b/aiosmtpd/__init__.py
@@ -12,7 +12,7 @@ def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
         warnings.simplefilter("ignore")
         try:
             loop = asyncio.get_event_loop()
-        except (DeprecationWarning, RuntimeError):
+        except (DeprecationWarning, RuntimeError):  # pragma: no cover
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
     return loop
--- a/aiosmtpd/__init__.py
+++ b/aiosmtpd/__init__.py
@@ -9,7 +9,7 @@ __version__ = "1.4.4a1"
 
 def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
     with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
+        warnings.simplefilter("error")
         try:
             loop = asyncio.get_event_loop()
         except (DeprecationWarning, RuntimeError):  # pragma: no cover
--- a/aiosmtpd/__init__.py
+++ b/aiosmtpd/__init__.py
@@ -8,11 +8,14 @@ __version__ = "1.4.4a1"
 
 
 def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
+    loop = None
     with warnings.catch_warnings():
         warnings.simplefilter("error")
         try:
             loop = asyncio.get_event_loop()
         except (DeprecationWarning, RuntimeError):  # pragma: py-lt-310
-            loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(loop)
+            if loop is None:  # pragma: py-lt-312
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+    assert isinstance(loop, asyncio.AbstractEventLoop)
     return loop
--- a/aiosmtpd/docs/NEWS.rst
+++ b/aiosmtpd/docs/NEWS.rst
@@ -16,10 +16,12 @@ Fixed/Improved
 * A whole bunch of annotations
 
 
-1.4.4a0 (ad hoc)
+1.4.4a1 (ad hoc)
 ================
 
-(Stub ``NEWS.rst`` entry as placeholder for ``qa`` test.)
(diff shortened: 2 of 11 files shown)


aiosmtpd/__init__.py
```diff
@@ -1,4 +1,18 @@
...
+
+__version__ = "1.4.4a1"
+
+
+def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        try:
+            loop = asyncio.get_event_loop()
+        except (DeprecationWarning, RuntimeError):  # pragma: no cover
```
pepoluan, line 15: When 3.12 comes out we might have to revisit this and change the caught exception in accordance to the actual exception that will be raised.

strongholdmedia:
Is there any sensible write-up as per why (and perhaps when) this policy stuff was introduced with asyncio?

pepoluan:
> Is there any sensible write-up as per why (and perhaps when) this policy stuff was introduced with asyncio?

The deprecation notice first appeared with Python 3.10: https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop

There's a discussion about this here: https://github.com/python/cpython/issues/83710

What's being discussed is over my head atm, though. So I'm just taking the 'easy way out' trying to ensure compat by implementing a workaround. And one ugly warning suppression.

pepoluan:
I hope @asvetlov can chime in here. I'm not sure where things are going, and I'd like this PR to be both backward- and forward-compatible with Python >= 3.7

The current changes indeed pass green when fed through the GHCI testing fleet, but it's still making me uneasy.

waynew:
I've taken a quick glance at this and based on the discussion I'm pretty sure that we need to make a decision within aiosmtpd:

Do we want to take responsibility for the created event loop?

It's looking like historically that's caused problems for folks generally. I think in our case we have two different scenarios:

1. aiosmtpd is the top-level app/process and should go ahead and create the event loop. Based on what I'm seeing on that thread, that should happen via `.run()`.
2. aiosmtpd is used as a library (like I do in [orouboros](https://github.com/waynew/orouboros)), and the authors should manage the loop.

What that looks like, I think, is that we should use `loop or asyncio.get_running_loop()`.

Comments, questions, rebuttals?

pepoluan:
> I've taken a quick glance at this and based on the discussion I'm pretty sure that we need to make a decision within aiosmtpd:
> 
> Do we want to take responsibility for the created event loop?
> 
> It's looking like historically that's caused problems for folks generally. I think in our case we have two different scenarios:
> 
>     1. aiosmtpd is the top-level app/process and should go ahead and create the event loop. Based on what I'm seeing on that thread, that should happen via `.run()`.
> 
>     2. aiosmtpd is used as a library (like I do in [orouboros](https://github.com/waynew/orouboros)), and the authors should manage the loop.

I'm all for treating `aiosmtpd` as a pure library, and letting users of `aiosmtpd` to provide the loop.

However that will introduce some possibly-breaking changes, especially for users that currently uses `aiosmtpd` as-is, relying on one of the Controllers to spin up an event loop. Like for instance one of the reasons for creation of UnthreadedController which allows user to spin up a different process and let UnthreadedController sets things up in that child process with minimal prepping.

So if we want to to go "pure library" route -- which for the record I'm all for it -- that change will need to wait until at the very least 1.5.0 ... or even 1.6.0 or 2.0.0

This PR is just kind of a band-aid over the underlying problem. At the very least it's meant to provide compat for 3.7 <= Python <= 3.12

waynew:
It looks like get_running_loop exists in 3.7 https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop

We should be able to use that - and if we want to retain compatibility just catch RuntimeError and start our own event loop.

For the record, I'm entirely OK with having entrypoints (e.g. http.server) that will actually spin up loops, we just need to do that in a responsible way.

pepoluan:
> It looks like get_running_loop exists in 3.7 https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop
> 
> We should be able to use that - and if we want to retain compatibility just catch RuntimeError and start our own event loop.

There's a catch though:

> This function can only be called from a coroutine or a callback.

Some calls to `get_event_loop()` took place in the main thread, and there `get_running_loop()` doesn't work. I forgot exactly what happened when I tried that, but tests got wonky.

aiosmtpd/__init__.py
```diff
@@ -1,4 +1,21 @@
...
+
+
+def _get_or_new_eventloop() -> asyncio.AbstractEventLoop:
+    loop = None
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        try:
+            loop = asyncio.get_event_loop()
+        except (DeprecationWarning, RuntimeError):  # pragma: py-lt-310
+            if loop is None:  # pragma: py-lt-312
```
pepoluan, line 17: On Python 3.11 (and I believe also 3.10), though `get_event_loop()` raises a warning when there's no existing event loop, it does indeed return a newly-created event loop. This defensive check prevents (re-)creation of a new event loop in that situation.

On Python 3.12 the behavior likely changes, i.e., no implicit creation of event loop inside `get_event_loop()`. In that situation, `loop` will remain `None` and only in that case will we create a new event loop.

pepoluan:
Aaaaand it seems that the new Warning behavior is getting reverted?

https://github.com/python/cpython/pull/100412

Nevertheless I still want to merge this changes to Cover-Our-Posteriors, so to speak.

waynew:
Well, they're just deferred,  and will eventually fail. I only glanced at this, if you'd like me to do a more thorough review I can, otherwise I'm OK with this approach for now

pepoluan:
Yeah, this PR basically just put a wrapper around asyncio.get_event_loop() to maintain the behavior of the function up to Python 3.9. So just a band-aid to the deeper problems.

I'd say for long term, just like you mentioned, we need to take a step back and implement asyncio more properly. And in the process probably we can also handle Issue #334 by using the 'more modern' way of using asyncio.

pepoluan:
This PR is not really 'urgent', though I've heard some complaints about the DeprecationWarning.

I'll let this simmer a bit for several more days just to ensure that I haven't overlooked something...

aiosmtpd/__init__.py
```diff
@@ -1,4 +1,21 @@
...
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+    assert isinstance(loop, asyncio.AbstractEventLoop)
```
waynew, line 20: **comment** FWIW this is probably less useful, since assertions can be disabled. And everything will go sideways shortly after this function gets called anyway :joy:

waynew (approved):
I didn't dig to check for places this should have been used or anything, but generally I do like this, and don't see any problems.

aiosmtpd/__init__.py
pepoluan: It's mostly for type checking, not expected to do anything useful 😂

PR #358, opened 2022-12-29
Change tox behavior

pepoluan:
## What do these changes do?

I had just managed to b0rk my cloned repo _completely_ when I specify tox to run `pypy38-nocov`.

That is due to `envdir` not specified explicitly so `tox` decides to 'clean up' my cloned repo and build its environment _right there_.

After recloning, I found out that tox now will recreate environment if the envname changes, so `py37-nocov` and `py37-cov` (for instance) can no longer share same packages. Meaning that previous "common deps" list is useless.

So I decided to just specify a specific envdir for all envname (i.e., combination of Pyton version + what test), and use conditionals on the deps installation.

Not really. However developers might find that hard disk usage balloons. Use `housekeep.py superclean`, it's there for a reason 😉

## Related issue number

None

## Checklist

  - [x] Windows (10): `pypy38-{nocov,cov}`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,6 +12,8 @@ envdir =
     py310: {toxworkdir}/3.10
     py311: {toxworkdir}/3.11
     pypy3: {toxworkdir}/pypy3
+    pypy37: {toxworkdir}/pypy37
+    pypy38: {toxworkdir}/pypy38
     py: {toxworkdir}/py
 commands =
     python housekeep.py prep
@@ -50,6 +52,8 @@ setenv =
     py310: INTERP=py310
     py311: INTERP=py311
     pypy3: INTERP=pypy3
+    pypy37: INTERP=pypy37
+    pypy38: INTERP=pypy38
     py: INTERP=py
 passenv =
     PYTHON*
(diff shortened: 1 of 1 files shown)


pepoluan:
I'm squamerging this rather urgently because I don't want the calamity that befell me happen to others.

tox.ini
```diff
@@ -4,15 +4,7 @@ envlist = qa, static, docs, py{37,38,39,310,311,py3}-{nocov,cov,diffcov}
...
 [testenv]
-# One virtualenv per Python version
-envdir =
-    py37: {toxworkdir}/3.7
-    py38: {toxworkdir}/3.8
-    py39: {toxworkdir}/3.9
-    py310: {toxworkdir}/3.10
-    py311: {toxworkdir}/3.11
-    pypy3: {toxworkdir}/pypy3
-    py: {toxworkdir}/py
```
pepoluan, line 15: Previously, if you specify something not listed here (e.g., `pypy38-nocov`), tox gets a bit confused and will decide to wipe out the whole cloned repo and creates its dirs in your repo's location. It also wipes recursively the contents of `.git`, so you can't do a `reset --hard`.

waynew: Oof. Well, that find sucks but glad you were able to recover :+1:

pepoluan: Thankfully there was no unpushed commits when that happened, so it's a quick git clone and I recovered.

PR #359, opened 2022-12-29
Drop testing for PyPy3.6, add PyPy3.9

pepoluan:
## What do these changes do?

No longer test with PyPy3.6.

PyPy3.6 has been EOL-ed for quite some time, and not available on non-Ubuntu runners anyways.

None.

## Related issue number

None

## Checklist


waynew (approved):
Into the future 🚀

PR #363, opened 2023-01-13
Migrate tool settings from setup.cfg to pyproject.toml

pepoluan:
## What do these changes do?

Practically all still-maintained projects are migrating to have a Single Source of Truth for their builds: `pyproject.toml`

This is in accordance with the guidelines in [PEP518](https://peps.python.org/pep-0518/) and [PEP621](https://peps.python.org/pep-0621/).

I have ascertained that all tools `aiosmtpd` uses -- both as dependency or for development -- now have support for putting their configuration in `pyproject.toml` (some needs a bit of workaround but works well). This left us with an empty `setup.cfg` file, which I removed because it's no longer useful (all setup information is now in `pyproject.toml` as well).

None if user has the latest `pip` and `setuptools`.

## Related issue number

None.

## Checklist

  - [x] GitHub CI 


setup.cfg
```diff
@@ -1,66 +1,12 @@
...
 exclude =
     aiosmtpd/docs/_exts/*
     _dump/*
+    aiosmtpd.egg-info/*
+    build/*
+    dist/*
+    Lib/*
+    Scripts/*
 disable =
     not-supported-yet
```
pepoluan: These were temporary; in a later commit these lines are deleted.

.github/workflows/unit-testing-and-coverage.yml
```diff
@@ -53,7 +51,9 @@ jobs:
             "print(config['flake8_plugins']['deps']);"
           )
           pip install "flake8>=5.0.4" $(python -c "${grab_f8_plugins[*]}")
-          python -m flake8 aiosmtpd setup.py housekeep.py release.py
+          echo "Running flake8 ... there will be no output if no errors:"
+          python -m flake8 aiosmtpd housekeep.py release.py
+          echo "flake8 done."
```
pepoluan, line 56: Added just to ensure that flake8 actually runs. flake8's silence on no error sometimes make me wonder if it ran or not 😅

release.py
```diff
@@ -28,8 +28,12 @@
     f"dist/aiosmtpd-{ver_str}-py3-none-any.whl",
 ]
 
+RELEASE_TOOLKIT = ["setuptools", "wheel", "twine", "build"]
```
pepoluan, line 31: `build` is the new build system created by PyPA that supports PEP 517.

pepoluan (commented):
A few comments

.github/workflows/unit-testing-and-coverage.yml
```diff
@@ -38,7 +36,7 @@ jobs:
         # language=bash
         run: |
           python -m pip install --upgrade pip setuptools wheel
-          python setup.py develop
+          python -m pip install .
```
waynew, line 39: Should that be install -e .?

that's more equivalent to develop

.github/workflows/unit-testing-and-coverage.yml
```diff
@@ -110,7 +112,7 @@ jobs:
         # Test deps
         pip install colorama "coverage>=7.0.1" coverage[toml] "coverage-conditional-plugin>=0.5.0" packaging pytest pytest-cov pytest-mock
         # Package deps
-        python setup.py develop
+        python -m pip install .
```
waynew, line 115: Same question here, -e?

README.rst
```diff
@@ -118,11 +118,12 @@ You can also request help by submitting a question on StackOverflow.
...
 ========
 
-You can install this package in a virtual environment like so::
+You can install this package -- after cloning -- in a virtual environment like so::
 
     $ python3 -m venv /path/to/venv
     $ source /path/to/venv/bin/activate
-    $ python setup.py install
+    $ python -m pip install -U pip setuptools
+    $ python -m pip install .
```
waynew, line 126: I think here we should definitely recommend editable mode, if we're suggesting folks clone flr dev. Otherwise they can just install from pypi

pyproject.toml
```diff
@@ -1,7 +1,67 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Topic :: Communications :: Email :: Mail Transport Agents",
+    "Framework :: AsyncIO",
+]
+dependencies = [
+    "atpublic",
+    "attrs",
+    "typing-extensions ; python_version < '3.8'",
```
waynew, line 39: Are typing extensions require to run aiosmtpd?

waynew (changes requested):
Aaaawesome! A few bits of feedback - not sure if anything really _needs_ to change...

.github/workflows/unit-testing-and-coverage.yml
pepoluan: Maybe? Ahaha this "install using pyproject.toml" is so new to me 😅

README.rst
pepoluan: Same as before. Let me do some study / experiments.

pyproject.toml
pepoluan: iirc it's to implement Protocol, I think that's in the Tproxy code.

Not really sure if that's actually needed to just run ...

waynew: Assuming that it's not needed for running we can put it in an `extras_require` (yeah, not extra_requires) section. I'm not sure what the precise form is for that, tho.

waynew:
@pepoluan FYI I'm still super interested in this getting in. There is a merge conflict in `__init__.py`. I've (re)reviewed PRs newer and up to this one, and will probably review more tomorrow afternoon.

.github/workflows/unit-testing-and-coverage.yml
webknjaz: Yes, `-e` would be the direct equivalent.

.github/workflows/unit-testing-and-coverage.yml
webknjaz: FWIW it's quite common to wrap linters with the pre-commit.com too.

README.rst
webknjaz: @pepoluan feel free to ask me if you have questions about Python packaging. I maintain PyPUG.

release.py
webknjaz: Yes. You need `build` for making dists and `twine` for linting them.

webknjaz: Also, PyPI no longer supports GPG. It's possible to use Sigstore, though.

My PyPUG article has an example of that: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

pyproject.toml
webknjaz: If that were true, it should be here, not in extras. FWIW it's only used in tests: https://github.com/aio-libs/aiosmtpd/blob/abadbd41dabb81eda4fbb196e1dc590bcc908648/aiosmtpd/tests/test_handlers.py#L30. Judging by the lack of it in runtime code, it might be fine to drop it from the tests and deps.

pyproject.toml
```diff
@@ -1,7 +1,67 @@
...
 [build-system]
 requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
+# endregion
+
+# region #### setuptools settings ##########
+
+[tool.setuptools]
+zip-safe = false
```
webknjaz, line 54: Why is it unsafe?

release.py
webknjaz:
```suggestion
RELEASE_TOOLKIT = ["build", "twine"]
```

PR #370, opened 2023-02-24
Allow setting the user which we drop permissions to using suid.

alexmv:
## What do these changes do?

We may want a different user than `nobody` when dropping permissions -- for example, if we intend to connect to a PostgreSQL database which does `ident` authentication.

No behaviour change.

## Checklist


aiosmtpd/main.py
```diff
@@ -224,12 +233,13 @@ def main(args: Optional[Sequence[str]] = None) -> None:
...
         try:
-            os.setuid(nobody)
+            os.setuid(suid_user_id)
         except PermissionError:
             print(
-                'Cannot setuid "nobody"; try running with -n option.', file=sys.stderr
+                f'Cannot setuid to "{args.suid_user}"; try running with -n option.',
+                file=sys.stderr,
             )
             sys.exit(1)
```
waynew, line 244: **suggestion (non-blocking)** Since we're going ahead and exiting with status 1 we *could* go ahead and simply `sys.exit(f"Cannot ...")` -- and use the (documented) functionality of sys.exit. However one of our tests *does* check that the exception `.value.code == 1`, and it would have to equal the error message instead. 

Not a big deal, but that would simplify things here just a bit.

waynew (changes requested):
Just a small change in regards to the documentation. And another optional change.

Otherwise, I'm +1 on this. I went ahead and checked it out locally and verified that the tests fail when the code is as-is on the main branch.

I do have a slight concern that there might be a better way to have the other connections made and then drop privileges to `nobody`... but I don't think that's necessarily a blocker here.

aiosmtpd/main.py
```diff
@@ -48,11 +48,20 @@ def _parser() -> ArgumentParser:
         default=True,
         action="store_false",
         help=(
-            "This program generally tries to setuid ``nobody``, unless this "
+            "This program generally uses setuid to drop permissions, unless this "
```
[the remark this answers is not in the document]
alexmv: Very true point!  Will repush with this adjustment.

aiosmtpd/main.py
alexmv: This feels very tangential to this PR.  Personally, I feel like the explicit print and exit code is more readable -- I've never seen the `sys.exit("...")` formulation, though it's by no means new!

So I'm not going to include this change in the PR, because I'm not entirely sold on it.

alexmv:
Thanks for the review!  I've repushed with your documentation change.

> I do have a slight concern that there might be a better way to have the other connections made and then drop privileges to nobody... but I don't think that's necessarily a blocker here.

I don't think there is -- neither `root` nor `nobody` can auth, so both making the connections before or after calling `suid(nobody)` isn't feasible.

PR #380, opened 2023-07-09
starttls: Clear unencrypted commands from buffer

Arusekk:
## What do these changes do?

When a STARTTLS command is not the last one in a pipeline, extra unencrypted commands may treated as part of the encrypted communication, effectively allowing to bypass the `ALLOWED_BEFORE_STARTTLS` list used when `require_starttls=True`.  This PR adds a buffer flush to ensure it does not happen.

No behaviour change.  Documentation reflects the changes already.

## Related issue number

None.

## Checklist

  - [x] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py36,py37,py38,py39}-{nocov,cov,diffcov}, qa, docs`

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -506,7 +506,7 @@ class SMTP(asyncio.StreamReaderProtocol):
             self.transport = transport
             # Discard any leftover unencrypted data
             # See https://tools.ietf.org/html/rfc3207#page-7
-            self._reader._buffer.clear()
+            self._reader._buffer.clear()  # type: ignore[attr-defined]
             # Do SSL certificate checking as rfc3207 part 4.1 says.  Why is
             # _extra a protected attribute?
             assert self._tls_protocol is not None
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -504,6 +504,9 @@ class SMTP(asyncio.StreamReaderProtocol):
             self._reader._transport = transport  # type: ignore[attr-defined]
             self._writer._transport = transport  # type: ignore[attr-defined]
             self.transport = transport
+            # Discard any leftover unencrypted data
+            # See https://tools.ietf.org/html/rfc3207#page-7
+            self._reader._buffer.clear()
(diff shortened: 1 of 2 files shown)


aiosmtpd/smtp.py
```diff
@@ -907,6 +907,11 @@ async def smtp_STARTTLS(self, arg: str):
             await self.push('454 TLS not available')
             return
         await self.push('220 Ready to start TLS')
+
+        # Discard any leftover unencrypted data
+        # See https://tools.ietf.org/html/rfc3207#page-7
+        self._reader._buffer.clear()
```
Dreamsorcerer, line 913: This isn't very robust as it depends on internal implementation details. Is there a beter solution we could try?

Arusekk: I thought this class does not care very much about StreamReader API stability, since it already hacks on the old StreamReader instead of re-creating it, when it sets `self._reader._transport = transport` in `connection_made` (StreamReader.set_transport does not support setting the transport again once it has been set for a reason probably).

This overwriting of an internal field is not a supported use, and therefore we must make for it by doing another hacky thing, which means clearing the buffer manually. This hack can alternatively be moved towards that hack, in order to be more apparent. WDYT?

Dreamsorcerer: Ah, I see. Yes, I think it'd make sense to move the code just after those lines to make it clearer.

Arusekk: Done.

aiosmtpd/smtp.py
```diff
@@ -504,6 +504,9 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
             self._reader._transport = transport  # type: ignore[attr-defined]
             self._writer._transport = transport  # type: ignore[attr-defined]
             self.transport = transport
+            # Discard any leftover unencrypted data
+            # See https://tools.ietf.org/html/rfc3207#page-7
+            self._reader._buffer.clear()
```
Dreamsorcerer, line 509:
```suggestion
            self._reader._buffer.clear()  # type: ignore[attr-defined]
```

PR #384, opened 2023-10-12
Small docs change

### LINKED ISSUE #320: docs update for 1.5
There's probably a lot of docs that we need to update for 1.5 - if anyone can go through and audit the docs for stuff like, old Python versions, etc. and add those here that would be awesome!

waynew:
See #320 - I did a quick audit and changed the couple of things that could use it. I did *not* confirm whether or not any of the code samples in the docs have any changes that are inconsistent with the current codebase... My hope is that they were already update and tested :sweat_smile: 

Anyway... this may or may not completely resolve #320.

### DIFF
--- a/aiosmtpd/docs/intro.rst
+++ b/aiosmtpd/docs/intro.rst
@@ -11,8 +11,7 @@ Local Mail Transfer Protocol (LMTP).  It is derived from
 standard library module, and provides both a command line interface and an API
 for use in testing applications that send email. Since the `smtpd module was
 removed in Python 3.12
-<https://docs.python.org/3.12/whatsnew/3.12.html#locale>`, following `PEP 594
-<https://peps.python.org/pep-0594/#smtpd>`, aiosmtpd is the recommended
+<https://docs.python.org/3.12/whatsnew/3.12.html#locale>`__, following :pep:`594#smtpd`, aiosmtpd is the recommended
 replacement.
 
 Inspiration for this library comes from several other packages:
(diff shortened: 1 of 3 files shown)


aiosmtpd/docs/intro.rst
pepoluan, line 1: I would split the piece of text explaining that we're replacement for `smtpd` into a separate paragraph, optionally with some kind of highlight / callout.

pepoluan (commented):
A thot about that ... 🙃

aiosmtpd/docs/intro.rst
```diff
@@ -9,7 +9,11 @@ Simple Mail Transfer Protocol (SMTP) and
 Local Mail Transfer Protocol (LMTP).  It is derived from
 `Python 3.5's smtpd.py <https://hg.python.org/cpython/file/3.5/Lib/smtpd.py>`__
 standard library module, and provides both a command line interface and an API
-for use in testing applications that send email.
+for use in testing applications that send email. Since the `smtpd module was
+removed in Python 3.12
+<https://docs.python.org/3.12/whatsnew/3.12.html#locale>`, following `PEP 594
+<https://peps.python.org/pep-0594/#smtpd>`, aiosmtpd is the recommended
```
webknjaz, line 15: External references need trailing underscores in RST. The current syntax uses an anonymous role. Also, for PEPs, there's a dedicated role `:pep:`:
```suggestion
<https://docs.python.org/3.12/whatsnew/3.12.html#locale>`__, following :pep:`594#smtpd`, aiosmtpd is the recommended
```

webknjaz:
I enabled building PR previews @ RTD but it fails currently because the RTD config file needs to be updated: https://readthedocs.org/projects/aiosmtpd/builds/23289246/.

The GHA linting is failing is unrelated either, because of the unpinned flake8 plugin version, I suppose.

webknjaz:
> the RTD config file needs to be updated

Filed an issue for this with pointers: https://github.com/aio-libs/aiosmtpd/issues/393.

webknjaz:
@waynew needs conflict resolution

PR #415, opened 2024-03-02
Release v1.4.5

Dreamsorcerer:


Dreamsorcerer:
@The-Login It appears your change has broken a test, could you investigate? Also, the NEWLINE global you updated appears to be unused, maybe it should be removed or the other places you changed could use that variable?

The-Login:
I've updated the SMTP smuggling test in https://github.com/aio-libs/aiosmtpd/pull/416.
Also, I removed the global NEWLINE variable in https://github.com/aio-libs/aiosmtpd/pull/417.

PR #473, opened 2024-06-28
Add compatibility for Python 3.13

rominf:
## What do these changes do?

- Fix tox for Python >= 3.12
- Remove `pytest-profiling`
- Remove `pytest-asyncio` traces
- Add compatibility for Python 3.13

## Are there changes in behavior for the user?

* Add compatibility for Python 3.13

## Related issue number

Closes #403.

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] tox testenvs have been executed in the following environments:
  
  - [x] Linux (Fedora 40): `{py38,py39,py310,py311,py312,py313}-{cov}`
- [x] Documentation reflects the changes
- [x] Add a news fragment into the `NEWS.rst` file
  
  * Add under the "aiosmtpd-next" section, creating one if necessary
    * You may create subsections to group the changes, if you like
  * Use full sentences with correct case and punctuation
  * Refer to relevant Issue if applicable

Issue #385, opened 2023-10-19
Replace flake8 test with ruff


pepoluan:
It's faster and natively supports `pyproject.toml` ... what's not to like?

ZeyadMoustafaKamal:
I think Ruff is a formater but flake8 is a linter. The main difference is that the formater fixes the code but the linter just tells you where is the problem. For me, I love to use flake8 only without black, ruff, or anything else. It's a personal recommendation only but I feel like they don't need it at all.

Another thing to keep in mind is that they look like to stopped maintaining this project as you see the last commit is from a year ago so I don't think they will do anything about it at all.

webknjaz:
Ruff is both linter and formatter. I believe, it's possible to opt into using just the linting mode. It's fast, but doesn't reimplement rules for every flake8 plugin. So my recommendation is to use Ruff in the linting mode and use flake8 additionally but only for rules that haven't been ported.

As for formatting, I personally dislike black but don't reject formatting as a concept. It tends to give a false impression of quality if overused. OTOH, that's up to the maintainers, of course.

And regarding the maintenance, it's done by volunteers who often step up to do something that they themselves need from the project. Recognize that they are human and don't necessarily have resources to provide free labor to random people on the internet.

Using the commit date or any similar metadata to define the maintenance status is just silly — if a project does what the maintainers need it to do, it doesn't really need as many updates. And accepting+reviewing random patches requires significant time commitments and you can't just bully them into doing this by declaring a project unmaintained.
It's open source, anyway — anybody is free to fork it and is allowed to develop such a fork as they see fit (within the limits of the licensing, of course).

If you want to join the maintenance, do mention that @ https://github.com/aio-libs/aiosmtpd/issues/144.

ZeyadMoustafaKamal:
This is the first time for me to hear that Ruff is a formatter and linter at the same time. Anyway, I don't think that it will add something to the project at all as from my experience, using one linter is enough. But if you want to use Ruff instead of flake8 you can wait for some time for it they add all the rules of flake8. 

And for the performance, I don't think it matters. I mean imagine that flake8 takes 1 second and ruff takes 0.01 second. even with this huge difference, I don't think that waiting for a second or waiting for a 0.01 second is a big thing that might make you move from flake8 to Ruff

[...]

PR #552, opened 2025-04-06
Use trustme for test TLS certificate and key

rominf:
## What do these changes do?

Use trustme for test TLS certificate and key

With the current approach TLS certificate and key are hardcoded in separate files and distributed as part of wheels. This is unnecessary for end users of the package, also this brings difficulties while supporting Python 3.13, see: https://github.com/aio-libs/aiosmtpd/pull/473

Use trustme for generating TLS certificate and key on the fly, as it is done in other aio-libs packages.

Initially proposed in: https://github.com/aio-libs/aiosmtpd/pull/473#discussion_r1745661940

Wheels no longer include test TLS certificate and key.

## Checklist

  - [x] Linux (openSUSE Tumbleweed): `py311-cov` (after merging https://github.com/aio-libs/aiosmtpd/pull/554)
- [ ] Documentation reflects the changes [not worthy for documentation]
- [ ] Add a news fragment into the `NEWS.rst` file [not worthy for NEWS.rst]
  

webknjaz:
It's unclear to me what the pytest-asyncio plug-in has to do with TLS certificates or why it needs to be removed. Either way, it's best to avoid mixing up unrelated changes.

rominf:
@webknjaz Unfortunately, `tox` tests did not pass on my machine, so I had to make this change. However, I understand your point about unrelated changes, so I split this PR into two. See: https://github.com/aio-libs/aiosmtpd/pull/554.

aiosmtpd/tests/conftest.py
```diff
@@ -315,27 +321,52 @@ def client(request: pytest.FixtureRequest) -> Generator[SMTPClient, None, None]:
...
+    if not TRUSTME:
+        pytest.xfail("trustme is not supported")
+    return trustme.CA()
+
+
+@pytest.fixture
+def tls_certificate(tls_certificate_authority: trustme.CA) -> trustme.LeafCert:
+    return tls_certificate_authority.issue_cert(
+        "localhost",
+        "xn--prklad-4va.localhost",
```
webknjaz, line 334: Let's log the decoded value here
```suggestion
        "xn--prklad-4va.localhost",  # príklad.localhost
```

aiosmtpd/tests/conftest.py
```diff
@@ -26,14 +26,22 @@
     _ProactorBasePipeTransport = None
     HAS_PROACTOR = False
 
+try:
+    import trustme
+
+    # Check if the CA is available in runtime, MacOS on Py3.10 fails somehow
```
webknjaz, line 32: I don't know if this is still true. Let's drop the check and just keep the plain import.

requirements.txt
```diff
@@ -5,3 +5,4 @@ pytest==8.3.4
 pytest-asyncio==0.24.0
 pytest-cov==5.0.0
 pytest-mock==3.14.0
+trustme==1.2.1
```
rominf, line 8: Latest version of `trustme` is not available for Python 3.8. Here is a PR to drop Python 3.8, as it's EOL: https://github.com/aio-libs/aiosmtpd/pull/555

rominf:
I tried to add workaround for installing `trustme`: see https://github.com/aio-libs/aiosmtpd/pull/552#discussion_r2041036210 and https://github.com/aio-libs/aiosmtpd/pull/552/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R8-R9, however it didn't work because of PyPy trying to build `cryptography`:
```
error: the configured PyPy interpreter version (3.8) is lower than PyO3's minimum supported version (3.9)
```

Hence, https://github.com/aio-libs/aiosmtpd/pull/555 should be reviewed and merged first.

webknjaz:
I'm pretty sure cryptography has older published wheels that would be compatible.

rominf:
> I'm pretty sure cryptography has older published wheels that would be compatible.

You are right, `cryptography==41.0.7 ` has compatible wheels published. I updated my PR. However, this version has a few security issues: https://security.snyk.io/package/pip/cryptography/41.0.7, so dependabot will likely create PRs.

webknjaz:
It was rejected and superseded by PEP 508.

webknjaz:
@rominf could you rebase and resolve the conflicts?

rominf:
@webknjaz Done. Please have a look at https://github.com/aio-libs/aiosmtpd/pull/589, which fixes the CI build for this PR.

webknjaz:
The dep no longer needs to be conditional w/ py38 having been dropped.

webknjaz:
Also the iterator import should come from the collections module, IIRC.

rominf:
> Also the iterator import should come from the collections module, IIRC.

@webknjaz, you could add ruff and the UP035 ruff rule to catch this automatically:

```
[tool.ruff.lint]
select = ["UP035"]
```

It enforces importing Iterator (and other deprecated typing aliases) from collections.abc instead of typing.

rominf:
@webknjaz The PR is ready for a review, all tests pass.

webknjaz (approved):
Thanks!

webknjaz:
If you'd like to try integrating Ruff, open a tracking issue for planning and pointers. I have prior art that would need to be followed.

loqs:
> If you'd like to try integrating Ruff, open a tracking issue for planning and pointers.

@webknjaz there is already https://github.com/aio-libs/aiosmtpd/issues/385. As a quicker measure flake8-black could be added to the github CI to detect formatting violations.

PR #567, opened 2026-03-23
Python 3.16 compat: use `inspect.iscoroutinefunction()`

loqs:
## What do these changes do?

Use inspect.iscoroutinefunction to replace asyncio.iscoroutinefunction deprecated in Python 3.14 and scheduled for deletion in Python 3.16. inspect.iscoroutinefunction has been available since Python 3.9.

No.
## Checklist

### DIFF
--- a/aiosmtpd/smtp.py
+++ b/aiosmtpd/smtp.py
@@ -1526,7 +1526,7 @@ class SMTP(asyncio.StreamReaderProtocol):
                 assert self.session is not None
                 args = (self.session.peer, self.envelope.mail_from,
                         self.envelope.rcpt_tos, self.envelope.content)
-                if asyncio.iscoroutinefunction(
+                if inspect.iscoroutinefunction(
                         self.event_handler.process_message):
                     status = await self.event_handler.process_message(*args)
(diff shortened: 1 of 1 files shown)


loqs:
As the project's tox is currently broken I ran pytest in Python 3.9 to 3.14 venv's instead.

According to https://github.com/aio-libs/aiosmtpd/blob/master/aiosmtpd/smtp.py#L1534-L1536 `iscoroutinefunction` is only used a deprecated API that was due to be removed in 1.0? An alternative would be to drop the API.

webknjaz:
> According to [`master`/aiosmtpd/smtp.py#L1534-L1536](https://github.com/aio-libs/aiosmtpd/blob/master/aiosmtpd/smtp.py?rgh-link-date=2026-03-23T20%3A56%3A05Z#L1534-L1536) `iscoroutinefunction` is only used a deprecated API that was due to be removed in 1.0? An alternative would be to drop the API.

I think the API removal should be a separate PR. Also, I'd like to have somebody with long-term maintenance intentions to make such a decision.

Issue #544, opened 2025-03-10
Outdated build instructions in README


cuu508:
I followed the instructions in the "BUILDING" section in README:

```
git clone git@github.com:aio-libs/aiosmtpd.git
python3 -m venv .venv
source .venv/bin/activate
python setup.py install
```

This produces:

```
Traceback (most recent call last):
  File "/tmp/aiosmtpd/setup.py", line 1, in <module>
    from setuptools import setup
ModuleNotFoundError: No module named 'setuptools'
```

Is there perhaps an instruction missing there?

I'm using Ubuntu 24.04 and system's default python, python 3.12.3.

cuu508:
I then tried to run the test suite. I installed and ran tox:

```
pip install tox
tox
```

This produces long output with various error messages. Here's the trailing part of it:


```
[...]
housekeep.py:205:37: ANN204 Missing return type annotation for special method
housekeep.py:210:22: ANN101 Missing type annotation for self in method
housekeep.py:210:29: ANN002 Missing type annotation for *args
housekeep.py:210:37: ANN003 Missing type annotation for **kwargs
housekeep.py:210:44: ANN204 Missing return type annotation for special method
housekeep.py:245:28: ANN201 Missing return type annotation for public function
qa: exit 1 (4.91 seconds) /tmp/aiosmtpd> python -m flake8 aiosmtpd setup.py housekeep.py release.py pid=2275641
static: skipped because platform linux does not match ^(?!win32)(?!cygwin)
qa: FAIL ✖ in 14.98 seconds
static: SKIP ⚠ in 0 seconds
docs: failed with Could not open requirements file /tmp/aiosmtpd/aiosmtpd/docs/RTD-requirements.txt: [Errno 2] No such file or directory: '/tmp/aiosmtpd/aiosmtpd/docs/RTD-requirements.txt' for tox env py within deps
docs: FAIL ✖ in 0.05 seconds
py38-nocov: skipped because could not find python interpreter with spec(s): py38
py38-nocov: SKIP ⚠ in 0 seconds
py38-cov: skipped because could not find python interpreter with spec(s): py38
py38-cov: SKIP ⚠ in 0 seconds
py38-diffcov: skipped because could not find python interpreter with spec(s): py38
py38-diffcov: SKIP ⚠ in 0 seconds
py39-nocov: skipped because could not find python interpreter with spec(s): py39
py39-nocov: SKIP ⚠ in 0 seconds
[...]

PR #569, opened 2026-04-10
Fix tox diffcover

loqs:
## What do these changes do?
Fix tox diffcover failure by:
- set cov-report output location to the location expected by diffcover.
- use --compare-branch master for diff-cover to match the project's branch.

No.

## Related issue number
Related https://github.com/aio-libs/aiosmtpd/issues/544.

## Checklist

  - [x] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py36,py37,py38,py39}-diffcov`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ commands =
     # So, run only if "not (310 or diffcov)" ==> "(not 310) and (not diffcov)"
     !py310-!diffcov: bandit -c bandit.yml -r aiosmtpd
     nocov: pytest --verbose -p no:cov --tb=short {posargs}
-    cov: pytest --cov --cov-report=xml --cov-report=html --cov-report=term --tb=short {posargs}
+    cov: pytest --cov --cov-report=xml:_dump/coverage-{env:INTERP}.xml --cov-report=html --cov-report=term --tb=short {posargs}
     diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --html-report htmlcov/diffcov-{env:INTERP}.html
     diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --fail-under=100
     profile: pytest --profile {posargs}
--- a/tox.ini
+++ b/tox.ini
@@ -12,8 +12,8 @@ commands =
     !py310-!diffcov: bandit -c bandit.yml -r aiosmtpd
     nocov: pytest --verbose -p no:cov --tb=short {posargs}
     cov: pytest --cov --cov-report=xml:_dump/coverage-{env:INTERP}.xml --cov-report=html --cov-report=term --tb=short {posargs}
-    diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --html-report htmlcov/diffcov-{env:INTERP}.html
-    diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --fail-under=100
+    diffcov: diff-cover --compare-branch master _dump/coverage-{env:INTERP}.xml --html-report htmlcov/diffcov-{env:INTERP}.html
+    diffcov: diff-cover --compare-branch master _dump/coverage-{env:INTERP}.xml --fail-under=100
     profile: pytest --profile {posargs}
     python housekeep.py --afterbar --afterbar gather
 #sitepackages = True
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -11,9 +11,9 @@ commands =
     # So, run only if "not (310 or diffcov)" ==> "(not 310) and (not diffcov)"
     !py310-!diffcov: bandit -c bandit.yml -r aiosmtpd
     nocov: pytest --verbose -p no:cov --tb=short {posargs}
-    cov: pytest --cov --cov-report=xml --cov-report=html --cov-report=term --tb=short {posargs}
-    diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --html-report htmlcov/diffcov-{env:INTERP}.html
-    diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --fail-under=100
+    cov: pytest --cov --cov-report=xml:_dump/coverage-{env:INTERP}.xml --cov-report=html --cov-report=term --tb=short {posargs}
```
webknjaz: Why don't we have two XML reports? Is there anything in the CI that expects the standard path?

loqs: > Why don't we have two XML reports?

`_dump/coverage-{env:INTERP}.xml` as set in pyproject.toml is ignored by both `python -m coverage xml` used by CI and `pytest --cov --cov-report=xml` used by tox. So only one report is generated.

> Is there anything in the CI that expects the standard path?

Yes the Codecov service.

`pytest --cov --cov-config=pyproject.toml --cov-report=xml` and `pytest --cov --cov-report=xml:_dump/coverage-${INTERP}.xml` both change the location to that expected by diffcover as run by tox.

The Codecov service can be adjusted to use `_dump/coverage-${INTERP}.xml` if you want to standardize that as the location.

webknjaz: I was thinking of having `--cov-report=xml --cov-report=xml:_dump/coverage-${INTERP}.xml` which would write both `coverage.xml` and `_dump/coverage-${INTERP}.xml`, IIUC.

---

In other projects, I've settled on having just these generic flags in
https://github.com/cherrypy/cheroot/blob/2ffb0baf8aa480da9da8d522b3fd7eab271f6dba/pytest.ini#L19-L26
in `pytest.ini`, while having local invocation-specific (contrib env) `{posargs}` with a fallback that produces HTML (https://github.com/cherrypy/cheroot/blob/2ffb0baf8aa480da9da8d522b3fd7eab271f6dba/tox.ini#L30C16-L30C58) but allows the CI (GHA) to pass args for making XML (coverage+junit: https://github.com/cherrypy/cheroot/blob/2ffb0baf8aa480da9da8d522b3fd7eab271f6dba/.github/workflows/ci-cd.yml#L804-L809).

aiosmtpd historically has `diff-cover` that I haven't looked into, and so I don't have a clear integration guidance. But long term, I'd like to migrate to what I outlined in https://github.com/aio-libs/aiosmtpd/issues/544#issuecomment-4205862595 and, if needed, integrate `diff-cover` in a way that would fit into that structure.

Right now, that's a bit out of the scope here. So if the following suggestion works for you, let's integrate it (and rebase the PR), and I'll be happy to merge:
```suggestion
    cov: pytest --cov --cov-report=xml --cov-report=xml:_dump/coverage-{env:INTERP}.xml --cov-report=html --cov-report=term --tb=short {posargs}
```

If this is somehow problematic, let me know.

loqs: Thank you for the suggestion. Implemented.

webknjaz:
There's 5 failed tests in the Windows job: https://github.com/aio-libs/aiosmtpd/actions/runs/25456153086/job/74686050166#step:5:589

This seems unrelated so I've restarted the CI. Though, it's probably worth investing and digging through the logs. If the tests are flaky, they should be stabilized.

webknjaz:
Note: I realized just now that the CI is not calling tox but invokes pytest and coverage directly, bypassing some shared configuration. I think it'd be good to start integrating `reusable-tox.yml` in a follow-up.

webknjaz (approved):

loqs:
> Though, it's probably worth investing and digging through the logs. If the tests are flaky, they should be stabilized.

I looked through the log. The failing tests are appear to Windows no allowing the same port to be reused in quick succession while this is not triggered on the Linux. I do not have a Windows system to debug the issue. I did notice some tests are already skipped on Windows.

loqs:
> Note: I realized just now that the CI is not calling tox but invokes pytest and coverage directly, bypassing some shared configuration. I think it'd be good to start integrating reusable-tox.yml in a follow-up.

I was planning on getting all current tox environments passing first. I estimate about five more pull requests four singles and a multiple commit one covering flake8 and hopefully they should all be passing.

PR #570, opened 2026-04-10
tox nocov: fix by overriding pytest.ini addopts

loqs:
## What do these changes do?

The nocov environments were failing because pytest.ini added in 8d81f96f3d91de995ecbcae01805f21a71410790 includes --cov options in addopts, which pytest treats as unrecognized arguments when the cov plugin is disabled with -p no:cov.

Adding --override-ini='addopts=' to the nocov command in tox.ini, clears the conflicting pytest.ini settings while keeping the GitHub CI workflow intact (which runs pytest directly without tox).

No.

## Related issue number

Related https://github.com/aio-libs/aiosmtpd/issues/544.

## Checklist

  - [x] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py36,py37,py38,py39}-nocov`
  
### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,7 @@ commands =
     # Bandit is not needed on diffcov, and seems to be incompatible with 310
     # So, run only if "not (310 or diffcov)" ==> "(not 310) and (not diffcov)"
     !py310-!diffcov: bandit -c bandit.yml -r aiosmtpd
-    nocov: pytest --verbose -p no:cov --tb=short {posargs}
+    nocov: pytest --verbose --no-cov --tb=short {posargs}
     cov: pytest --cov --cov-report=xml --cov-report=html --cov-report=term --tb=short {posargs}
     diffcov: diff-cover _dump/coverage-{env:INTERP}.xml --html-report htmlcov/diffcov-{env:INTERP}.html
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -10,7 +10,7 @@ commands =
     # Bandit is not needed on diffcov, and seems to be incompatible with 310
     # So, run only if "not (310 or diffcov)" ==> "(not 310) and (not diffcov)"
     !py310-!diffcov: bandit -c bandit.yml -r aiosmtpd
-    nocov: pytest --verbose -p no:cov --tb=short {posargs}
+    nocov: pytest --verbose -p no:cov --override-ini='addopts=' --tb=short {posargs}
```
webknjaz: The actual problem here is that the plugin is importable via `pytest_cov` and so preventing it from being loaded would be `-pno:pytest_cov`.

But this plugin has a `--no-cov` option, which is probably what was intended originally.

I'd probably do
```suggestion
    nocov: pytest --verbose --no-cov --tb=short {posargs}
```

and let the plugin be installed unconditionally. Though, this all likely needs a redesign.

loqs: I have implemented your suggestion.

webknjaz (approved):

PR #583, opened 2026-05-07
Remove use of ramdom.getrandbits.

loqs:
## What do these changes do?
Remove use of ramdom.getrandbits. The tests for random port use were duplicates. Other uses of getrandombits were converted to fixed ports.
No
## Related issue number
Related https://github.com/aio-libs/aiosmtpd/issues/544
## Checklist
  
  - [x] Linux (Ubuntu 18.04, Ubuntu 20.04, Arch): `{py39.py310,py311,py312}-{nocov,cov,diffcov}, qa, docs`

### DIFF
--- a/aiosmtpd/tests/test_proxyprotocol.py
+++ b/aiosmtpd/tests/test_proxyprotocol.py
@@ -5,13 +5,11 @@ import asyncio
 import errno
 import logging
 import operator
-import random
 import socket
 import struct
 import time
 from base64 import b64decode
 from contextlib import contextmanager, suppress
-from functools import partial
 from ipaddress import IPv4Address, IPv6Address
 from smtplib import SMTP as SMTPClient
 from smtplib import SMTPServerDisconnected
@@ -43,7 +41,6 @@ TIMEOUT_MULTIPLIER = 2.0
 
 param = pytest.param
 parametrize = pytest.mark.parametrize
-random_port = partial(random.getrandbits, 16)
 
 TEST_TLV_DATA_1 = (
     b"\x03\x00\x04Z\xfd\xc6\xff\x02\x00\tAUTHORITI\x05\x00\tUNIKUE_ID"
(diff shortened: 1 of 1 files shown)


loqs:
There are multiple alternatives if you prefer:
- Skip bandit for the tests directory
- Switch from random to secrets
- Stop using a random port as the port number is only used for mocking.

aiosmtpd/tests/test_proxyprotocol.py
```diff
@@ -518,8 +518,8 @@ def test_tcp4(self, setup_proxy_protocol):
 
     def test_tcp4_random(self, setup_proxy_protocol):
         setup_proxy_protocol(self)
-        srcip = ".".join(f"{random.getrandbits(8)}" for _ in range(0, 4))
-        dstip = ".".join(f"{random.getrandbits(8)}" for _ in range(0, 4))
+        srcip = ".".join(f"{random.getrandbits(8)}" for _ in range(0, 4))  # nosec: B311
```
webknjaz: Can bandit be configured to allow such things in tests? When it's wired through a flake8 plugin, I usually add that under `per-file-ignores`. Also, when ignoring inline, it's a good idea to attach a justification comment.

loqs: Do you have any preference between adding comments to inline ignores and wiring bandit into flake8 and using per-file-ignores?

loqs: After https://github.com/aio-libs/aiosmtpd/pull/588 is merged I will wire bandit into flake8 and use a per file ignore.

webknjaz:
> * Switch from random to secrets
> 
> * Stop using a random port as the port number is only used for mocking.

I think these two would be preferential over suppressions. Normally, I'd ask for the ephemeral port 0 in e2e tests. But if this is only for mocking, I don't see how randomizing is useful at all, and so I don't have a problem with hardcoding a few values like 666 and 999, for example. This sounds like the cleanest solution to me.

loqs:
> I think these two would be preferential over suppressions. Normally, I'd ask for the ephemeral port 0 in e2e tests. But if this is only for mocking, I don't see how randomizing is useful at all, and so I don't have a problem with hardcoding a few values like 666 and 999, for example. This sounds like the cleanest solution to me.

After making your proposed change it became clear the tests were duplicates. Confirmed by removing them and `cover`/`diffcover` reporting 100% coverage after the removal.

aiosmtpd/tests/test_proxyprotocol.py
cuu508, line 1: To me it also looks like `test_tcp4_random` and `test_tcp6_random` are redundant.

The hardcoded ports in `test_malformed_addr` are not a problem – the code under test does not bind to them, we're only testing IP and port parsing and validation. So there's no risk of port collision.

cuu508 (approved):

webknjaz (approved):
Okay, let's get this in. But in the future, it's probably a good idea to to attach code comments explaining why it's safe to hardcode ports.

PR #621, opened 2026-07-11
Fix the authenticated_relayer example so authentication can succeed

### LINKED ISSUE #475: examples/authenticated_relayer/server.py uses a random salt value
line 36
`hashpass = pbkdf2_hmac("sha256", password, secrets.token_bytes(), 1000000).hex()`
The third argument in this function generates a random salt. It should use the previously generated salt from make_user_db.py
Otherwise the resulting hash will always be different from the stored password hash.

rodrigobnogueira:
## What do these changes do?

Make the `examples/authenticated_relayer` example actually able to authenticate a user. Two defects made success impossible:

1. `make_user_db.py` hashed each password with `secrets.token_bytes()` but never stored the salt, and `server.py` verified the presented password with *another* fresh random salt, so the two hashes could never be equal (#475).
2. The username lookup passed the raw `bytes` from `LoginPassword` to sqlite, which never matches the `text` column the DB stores, so even the user row was never found.

Changes:

* `make_user_db.py`: store a per-user `salt` column (hex) next to the hash.
* `server.py`: fetch the stored salt, recompute the pbkdf2 hash with it, and compare via `hmac.compare_digest`; decode the username before the lookup.
* One NEWS.rst line under 1.4.7.

Verified end-to-end by generating the DB and driving `Authenticator` directly: correct password accepted; wrong password, unknown user, and unsupported mechanism rejected.

Fixes #475

### DIFF
--- a/aiosmtpd/tests/test_examples.py
+++ b/aiosmtpd/tests/test_examples.py
@@ -6,6 +6,7 @@
 import importlib.util
 import os
 import runpy
+import sqlite3
 import sys
 from pathlib import Path
 from types import ModuleType
@@ -139,8 +140,6 @@ def test_stale_database_rejects_without_leaking(
     every attempt; letting sqlite3.OperationalError propagate turns it into
     a 500 quoting the schema back at an unauthenticated client.
     """
-    import sqlite3
-
     stale = tmp_path / "stale.db"
     conn = sqlite3.connect(stale)
     conn.execute("CREATE TABLE userauth (username text, hashpass text)")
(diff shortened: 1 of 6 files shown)


Vizonex (approved):
I would try adding in some tests first. Otherwise this is off to a good start.

aiosmtpd/tests/test_examples.py
```diff
@@ -0,0 +1,227 @@
...
+    assert result.success is False
+    assert result.handled is False
+
+
+def test_non_utf8_username_rejected_not_crashing(authenticate: Authenticate) -> None:
+    result = authenticate(
+        "LOGIN", LoginPassword(b"\xff\xfe" + A_USER.encode("utf-8"), A_PASSWORD)
+    )
+    assert result.success is False
+    assert result.handled is False
```
webknjaz, line 132: these look like the same test to me

aiosmtpd/tests/test_examples.py
```diff
@@ -0,0 +1,227 @@
...
+
+import aiosmtpd
+from aiosmtpd.smtp import AuthResult, LoginPassword
+
+# The examples are not part of the installed package; they only exist in a
+# repo checkout.  Skip cleanly when running from an installed aiosmtpd.
+_REPO_ROOT = Path(aiosmtpd.__file__).resolve().parent.parent
+RELAYER_DIR = _REPO_ROOT / "examples" / "authenticated_relayer"
+if not (RELAYER_DIR / "server.py").exists():
+    pytest.skip(
```
webknjaz, line 24: Couldn't this be an `importorskip` too then?

aiosmtpd/tests/test_examples.py
```diff
@@ -0,0 +1,227 @@
...
+    spec = importlib.util.spec_from_file_location(name, RELAYER_DIR / filename)
+    assert spec is not None
+    assert spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+make_user_db = _load_example("authrelay_make_user_db", "make_user_db.py")
```
webknjaz, line 44: This could've been an `importorskip` in a fixture injected into a test that would integrate better, than import-level loading I think. Plus it's easier for static analysis tooling to make sense of.

aiosmtpd/tests/test_examples.py
```diff
@@ -0,0 +1,227 @@
...
+    tmp_path: Path,
+) -> None:
+    """Decoding with errors="replace" would let one login match another user.
+
+    Every invalid byte becomes U+FFFD, so a login of b"\\xff" would match a
+    stored user named U+FFFD. Strict decoding rejects it instead.
+    """
+    salt = bytes(32)
+    db = tmp_path / "replacement.db"
+    conn = sqlite3.connect(db)
```
webknjaz, line 208: Doesn't this support a context manager interface?

aiosmtpd/tests/test_examples.py
loqs: Collapse them to one one parametrized test?

aiosmtpd/tests/test_examples.py
```diff
@@ -0,0 +1,227 @@
...
+    """Build the user DB once, with the real make_user_db.py script."""
+    db_dir = tmp_path_factory.mktemp("authrelay")
+    # Run it the way the example tells you to, in its own process, so the
+    # suite never has to chdir out from under the tests that follow.
+    subprocess.run(
+        [sys.executable, str(RELAYER_DIR / "make_user_db.py")],
+        cwd=db_dir,
+        check=True,
+        capture_output=True,
+    )
```
loqs, line 65: This triggers a warning from bandit:
```
>> Issue: [B603:subprocess_without_shell_equals_true] subprocess call - check for execution of untrusted input.
   Severity: Low   Confidence: High
   CWE: CWE-78 (https://cwe.mitre.org/data/definitions/78.html)
   More Info: https://bandit.readthedocs.io/en/1.9.4/plugins/b603_subprocess_without_shell_equals_true.html
   Location: aiosmtpd/tests/test_examples.py:60:4
59	    # suite never has to chdir out from under the tests that follow.
60	    subprocess.run(
61	        [sys.executable, str(RELAYER_DIR / "make_user_db.py")],
62	        cwd=db_dir,
63	        check=True,
64	        capture_output=True,
65	    )
66	    return db_dir / make_user_db.DB_FILE
```
Add an annotation to suppress?

aiosmtpd/tests/test_examples.py
loqs: It does however I believe it only covers transactions and not closing the connection.

Issue #613, opened 2026-06-12
Is release.py still used?


loqs:
Replacing direct use of bandit with flake8-bandit causes `housekeep.py` to be scanned for security issues:
```
flake8 aiosmtpd housekeep.py release.py setup.py --select S
pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
Unable to find qualified name for module: housekeep.py
Unable to find qualified name for module: release.py
Unable to find qualified name for module: setup.py
release.py:101:1: S603 subprocess call - check for execution of untrusted input.
release.py:103:1: S603 subprocess call - check for execution of untrusted input.
release.py:109:1: S603 subprocess call - check for execution of untrusted input.
release.py:126:1: S603 subprocess call - check for execution of untrusted input.
release.py:141:1: S603 subprocess call - check for execution of untrusted input.
release.py:156:1: S607 Starting a process with a partial executable path
release.py:156:1: S603 subprocess call - check for execution of untrusted input.
release.py:158:1: S607 Starting a process with a partial executable path
release.py:158:1: S603 subprocess call - check for execution of untrusted input.
```
If it is not used the simplest solution would be to delete it. Otherwise the detected issues need to be reviewed and addressed.

webknjaz:
Likely not. The previous process had GPG signing and interactivity. But currently, GHA seems to be set up to handle releases somewhat similarly to other projects. The last release used TP, meaning it used GHA. But perhaps, some parts of that script could be useful if moved to GHA.

Issue #612, opened 2026-06-04
flake8-bandit is listed in requirements-dev.txt but not selected


loqs:
bandit is also run as a standalone in tox.ini.
```
flake8 --select "S" aiosmtpd
aiosmtpd/__init__.py:20:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/controller.py:247:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/controller.py:257:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/controller.py:308:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/controller.py:355:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/controller.py:358:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/docs/_exts/autoprogramm.py:172:1: S102 Use of exec detected.
aiosmtpd/docs/_exts/autoprogramm.py:184:1: S307 Use of possibly insecure function - consider using safer ast.literal_eval.
aiosmtpd/docs/_exts/autoprogramm.py:311:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:108:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:109:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:128:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:137:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:141:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:142:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/handlers.py:203:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
aiosmtpd/proxy_protocol.py:488:1: S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
[...]

Issue #611, opened 2026-06-04
flake8 C90 is selected but max-complexity is not set


loqs:
```
flake8 --select "C90" aiosmtpd --isolated --max-complexity 10
aiosmtpd/docs/_exts/autoprogramm.py:70:1: C901 'scan_programs' is too complex (11)
aiosmtpd/docs/_exts/autoprogramm.py:204:5: C901 'AutoprogrammDirective.make_rst' is too complex (13)
aiosmtpd/main.py:170:1: C901 'parseargs' is too complex (13)
aiosmtpd/main.py:217:1: C901 'main' is too complex (12)
aiosmtpd/proxy_protocol.py:166:5: C901 'ProxyTLV.parse' is too complex (13)
aiosmtpd/proxy_protocol.py:343:1: C901 '_get_v1' is too complex (19)
aiosmtpd/proxy_protocol.py:427:1: C901 '_get_v2' is too complex (16)
aiosmtpd/smtp.py:314:5: C901 'SMTP.__init__' is too complex (18)
aiosmtpd/smtp.py:613:5: C901 'SMTP._handle_client' is too complex (33)
aiosmtpd/smtp.py:838:5: C901 'SMTP.smtp_EHLO' is too complex (13)
aiosmtpd/smtp.py:943:5: C901 'SMTP.smtp_AUTH' is too complex (16)
aiosmtpd/smtp.py:1269:5: C901 'SMTP.smtp_MAIL' is too complex (20)
aiosmtpd/smtp.py:1342:5: C901 'SMTP.smtp_RCPT' is too complex (13)
aiosmtpd/smtp.py:1408:5: C901 'SMTP.smtp_DATA' is too complex (28)
```

Issue #610, opened 2026-06-04
Consider dropping flake8 select usage


loqs:
Using flake8 select overrides plugins default selection of rules to enable. This means all plugins must be added to the select option and experimental and opinionated rules that the plugin's author may not enable by default have to be ignored.

webknjaz:
Yes, I use `extend-select` / `extend-ignore` in all my projects.

PR #622, opened 2026-07-11
Rationalize the flake8 configuration

rodrigobnogueira:
## What do these changes do?

Rationalize the flake8 configuration so that what is configured is what actually runs, resolving the nine related issues @loqs filed. The root problem was drift: tox's qa env and the GA lint job installed **different plugin subsets**, and the `select` list overrode plugin defaults, so several installed tools never ran anywhere.

Per issue:

- **#604** — the `[tool.isort]` block is removed from `pyproject.toml`: isort was never a dev dependency and its config conflicted with flake8-import-order (`pycharm` style), which is the tool now actually enforced. Import order across the package is fixed accordingly.
- **#605** — flake8-requirements now runs; the I900s it surfaced are resolved by declaring the genuinely-missing requirements (`colorama`, `packaging`, `docutils`) and mapping `atpublic:[public]` via `known-modules`. The optional guarded `typing_extensions` import gets a targeted `noqa`.
- **#606** — the dangling `enable-extensions = G` is removed rather than adopting flake8-logging-format: the plugin still imports `pkg_resources` at module level and cannot load under modern setuptools (this is what broke the first CI run of this PR). The G003/G004 call sites it would have flagged (f-strings and `+` in logging calls) are converted to lazy `%` formatting here anyway, so the cleanup loses nothing.
- **#607** — stale `SIM106` and `W293` ignores removed.
- **#608** — flake8-builtins now runs; the single A001 (`copyright` in `docs/conf.py`) is a per-file-ignore since that name is the Sphinx convention.
- **#609** — flake8-bugbear now runs, plus opt-in `B950` (the existing `E501` ignore finally has the backing it claimed); the five overlong lines it caught are wrapped.
- **#610** — `select`/`ignore` become `extend-select`/`extend-ignore`, so installed plugins contribute their default rule sets, as you suggested.
- **#611** — `max-complexity = 34` makes the long-selected-but-inert C90 actually run, as a no-regression ratchet over the current worst offender (`SMTP._handle_client`, 33). Verified: `--max-complexity 10` flags 14 functions, the configured bound flags none.
- **#612** — flake8-bandit now runs and the standalone bandit invocation in `[testenv]` is removed as duplicate; `bandit.yml` is retired with its `B101`/`B404` skips carried over as `S101`/`S404` ignores, and the S findings in maintainer tooling (`release.py`, qa suite) are per-file-ignored with reasons.

Structurally, `requirements-dev.txt` is now the **single plugin list**: the tox qa env consumes it via `-rrequirements-dev.txt` and the `[flake8_plugins]` pseudo-section is gone, so the qa env and the GA lint job can no longer drift apart.

One deliberate scope call: flake8-annotations and flake8-annotations-coverage are dropped rather than enforced. They never ran in CI, the codebase currently has 1,700+ ANN violations, and mypy is the typing gate; enforcing them is a separate epic (same spirit as the existing cognitive-complexity note). Also out of scope: #613 (release.py), which is a release-tooling decision, not lint config.

## Verification

- `python -m flake8 aiosmtpd setup.py housekeep.py release.py` (the exact CI invocation): **0 violations**, verified against setuptools 83 (no `pkg_resources`) so every pinned plugin actually loads in a current environment
- Teeth check: a scratch file with seeded violations correctly raises `A001 DUO104 G004 I100 S307`
- Each issue's repro command from its report re-run: clean under the new config
- Full test suite: **566 passed** (the G-fixes touch `smtp.py`/`main.py`/`proxy_protocol.py` runtime code)
- `mypy`: clean; `check-manifest`: clean; Sphinx HTML build (includes the NEWS entry): succeeds
- `pip install -r requirements-dev.txt` resolves (the docutils pin respects sphinx 8.2's `<0.22` bound)

I read the note on #604 about aligning configs with your preferred template (possibly Ruff) later. This PR intentionally stays within flake8 and implements the `extend-select`/`extend-ignore` shape you described on #610 - a coherent flake8 config should also make any later Ruff port mechanical. Happy to adjust or park it if you'd rather go straight to Ruff.


loqs:
This is intended to replace https://github.com/aio-libs/aiosmtpd/pull/588? I had a very similar series drafted as a follow on for 588.

rodrigobnogueira:
> This is intended to replace #588? I had a very similar series drafted as a follow on for 588.

Sorry, I missed #588 when opening this. It predates #604–#612 and references #544, so it never showed up in those issues' cross-references. Happy to reshape this PR if you want: I'd drop my flake8-annotations removal in favor of the restored ignores in #588 (which also has the real root cause — 8d81f96 removed them), and drop my test_smtp.py fix that duplicates yours, then rebase after #588 lands.
As far as I checked, the rest of this PR doesn't overlap with #588: the extend-select migration, the plugin-list unification between tox and requirements-dev.txt, the five newly-enforced plugins (bugbear, builtins, bandit, import-order, requirements), max-complexity, the standalone-bandit dedup, the isort-block removal, and the missing dep declarations.
WDYT?

loqs:
@rodrigobnogueira was I not working fast enough on the issues that were assigned to me?

rodrigobnogueira:
> @rodrigobnogueira was I not working fast enough on the issues that were assigned to me?

Not at all. This is entirely on me. Sorry, I didn't check the issue assignments, I only looked at the issues' cross-references, which is also how I missed #588.

PR #631, opened 2026-09-06
Fix authenticated relayer password verification

### LINKED ISSUE #475: examples/authenticated_relayer/server.py uses a random salt value

ryux1:
## What do these changes do?

Store each user's PBKDF2 salt in the authenticated relayer example database and use that salt when verifying a login. The authenticator decodes SMTP login bytes before querying SQLite, compares password digests with `hmac.compare_digest()`, and rejects incompatible legacy databases with regeneration guidance.

The example now propagates startup errors, remains active until shutdown, and cleans up its controller after cancellation or partial startup failure. Regression tests cover successful and failed authentication, the legacy schema, controller lifetime, and startup cleanup.

Yes. Credentials generated by `make_user_db.py` can now authenticate successfully. Existing databases using the old schema must be regenerated because their salts were never stored and cannot be recovered.

## Related issue number

Fixes #475

## Checklist

- [ ] I think the code is well written
- [x] Relevant test environments have been executed:
  - [x] Linux, Python 3.13: full suite, 570 passed
  - [x] Linux, Python 3.13: QA, 9 passed
  - [x] Linux, Python 3.13: MyPy and flake8 passed
- [x] Added a news entry to `NEWS.rst`

Drafted with OpenAI Codex; personal review by the contributor remains required before the PR is marked ready.

### DIFF
--- a/aiosmtpd/docs/NEWS.rst
+++ b/aiosmtpd/docs/NEWS.rst
@@ -13,7 +13,7 @@ Fixed/Improved
 * Dropped Python 3.8, PyPy 3.8
 * Added PyPy 3.11, dropped PyPy 3.9
 * Fixed the authenticated relayer example to verify passwords with their stored salts
-  (Closes #475)
+  and report incompatible databases at startup (Closes #475)
 
 
 1.4.6 (2024-05-18)
--- a/aiosmtpd/docs/NEWS.rst
+++ b/aiosmtpd/docs/NEWS.rst
@@ -12,6 +12,8 @@ Fixed/Improved
 
 * Dropped Python 3.8, PyPy 3.8
 * Added PyPy 3.11, dropped PyPy 3.9
+* Fixed the authenticated relayer example to verify passwords with their stored salts
+  (Closes #475)
 
 
 1.4.6 (2024-05-18)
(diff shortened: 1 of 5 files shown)


ryux1:
Addressed in b491d32. Since the legacy hashes were created with salts that were never persisted, they cannot be migrated or verified. Authenticator initialization now validates the required columns and fails with an instruction to recreate the database using make_user_db.py. A regression test covers the legacy two-column schema, and the full suite passes with 568 tests.

ryux1:
Addressed in 7643a49. The example now owns the main coroutine with asyncio.run(), so authenticator startup errors propagate and terminate the process. On successful startup, amain() waits until cancellation and stops the controller in its finally block instead of stopping it immediately. The new lifecycle test verifies that the controller remains running and is stopped on cancellation; the full suite passes with 569 tests.

ryux1:
Addressed in d783988. `Controller.start()` is now inside the `try/finally`, so a partial startup failure still invokes `stop()` before the exception propagates. A regression test uses a controller whose `start()` raises and verifies cleanup runs. Focused tests, flake8, and MyPy pass; the full suite now passes with 570 tests.

